PR #6, opened 2022-09-05
Fix twind theme generation on Windows

marialuizaaires:


lucis (approved):
LGTM!

Although, I feel we should switch to a more safe approach in the future

PR #7, opened 2022-09-05
Supabase client from token function

VerasThiago:
Small function to create supabase client in case user already have the token

### DIFF
--- a/supabase.ts
+++ b/supabase.ts
@@ -32,3 +32,9 @@ export function getSupabaseClientForUser(req: Request) {
   client.auth.setAuth(jwt);
   return client;
 }
+
+export function getSupabaseClientForUserFromToken(jwt: string) {
+  const client = createClient(userEndpoint, userKey);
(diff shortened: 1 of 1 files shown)


igorbrasileiro:
Silly question, why does this [function](https://github.com/deco-cx/live/pull/7/files#diff-d149b22a9ace8e6045e0fdd0fe9f857457f24644fe11507c867691ae4180030bR28-R34) doesn't work for you?

supabase.ts
```diff
@@ -32,3 +32,9 @@ export function getSupabaseClientForUser(req: Request) {
   client.auth.setAuth(jwt);
   return client;
 }
+
+export function getSupabaseClientForUserFromToken(jwt: string) {
```
guifromrio, line 36: TODO: rewrite getSupabaseClientForUser to use getSupabaseClientForUserFromToken

guifromrio (approved):

PR #10, opened 2022-09-13
feat: List Components Preview

igorbrasileiro:
### Context
The Live Editor, already has a way to add components, but the user doesn't have enough information about the component, because he/she only sees the component label. It would be good to have a component preview to improve the UX.

### Solution
To achieve this new UX, this PR adds a Component List Modal, that renders all components and islands from a project inside a modal. 

#### Features added:
- [x] Component Preview Modal
- [x] Route `/live/api/[components]` return a list of component preview info (islands and components) for components/islands that export const schema. This route is one year cached, and immutable if contains __fresh_c=<ID>
 The shape of the response:

```ts
interface ComponentPreview {
  link: string;
  component: string; // Relative path to components/islands
  componentLabel: string; // schema.title or component relative path.
}

interface Output { islands: ComponentPreview[], components: ComponentPreview[] }
```

- [x] Route `/live/api/[components]/<component-name>` to render components. This route is one year cached, and immutable if contains __fresh_c=<ID>

#### Changes:
- [x] LivePage handles children
- [x] Manifest schema has schemas from components that export the schema, otherwise, the component can't be there
- [x] Refactor manifest and liveOptions to LiveContext class, singleton

### DIFF
--- a/context.ts
+++ b/context.ts
@@ -9,6 +9,8 @@ let liveOptions: EnhancedLiveOptions;
 const defaultDomains = [
   `localhost`,
 ];
+const deploymentId = Deno.env.get("DENO_DEPLOYMENT_ID");
+const _isDenoDeploy = deploymentId !== undefined;
 
 export function getDefaultDomains() {
   return defaultDomains;
@@ -36,4 +38,12 @@ export function getLiveOptions() {
 
 export function setLiveOptions(newLiveOptions: EnhancedLiveOptions) {
   liveOptions = newLiveOptions;
+}
+
+export function getDeploymentId() {
+  return deploymentId;
+}
+
+export function isDenoDeploy() {
+  return _isDenoDeploy;
 }
--- a/context.ts
+++ b/context.ts
@@ -18,6 +18,10 @@ export function pushDefaultDomains(...domains: string[]) {
   defaultDomains.push(...domains);
 }
 
+export function isPrivateDomain(domain: string) {
+  return defaultDomains.includes(domain);
+}
+
 export function getManifest() {
   return manifest;
 }
--- a/context.ts
+++ b/context.ts
@@ -1,0 +1,35 @@
+import type { DecoManifest, LiveOptions } from "./types.ts";
+
+type EnhancedLiveOptions = LiveOptions & { siteId?: number };
+
+// While Fresh doesn't allow for injecting routes and middlewares,
+// we have to deliberately store the manifest in this scope.
+let manifest: DecoManifest;
+let liveOptions: EnhancedLiveOptions;
+const defaultDomains = [
+  `localhost`,
+];
+
+export function getDefaultDomains() {
+  return defaultDomains;
+}
+
+export function pushDefaultDomains(...domains: string[]) {
+  defaultDomains.push(...domains);
+}
+
+export function getManifest() {
+  return manifest;
+}
+
+export function setManifest(newManifest: DecoManifest) {
+  manifest = newManifest;
+}
+
+export function getLiveOptions() {
+  return liveOptions;
+}
+
+export function setLiveOptions(newLiveOptions: EnhancedLiveOptions) {
+  liveOptions = newLiveOptions;
+}
--- a/context.ts
+++ b/context.ts
@@ -2,48 +2,71 @@ import type { DecoManifest, LiveOptions } from "./types.ts";
 
 type EnhancedLiveOptions = LiveOptions & { siteId?: number };
 
-// While Fresh doesn't allow for injecting routes and middlewares,
-// we have to deliberately store the manifest in this scope.
-let manifest: DecoManifest;
-let liveOptions: EnhancedLiveOptions;
-const defaultDomains = [
-  `localhost`,
-];
-const deploymentId = Deno.env.get("DENO_DEPLOYMENT_ID");
-const _isDenoDeploy = deploymentId !== undefined;
+class LiveContext {
+  // While Fresh doesn't allow for injecting routes and middlewares,
+  // we have to deliberately store the manifest in this scope.
+  #manifest: DecoManifest;
+  #liveOptions: EnhancedLiveOptions;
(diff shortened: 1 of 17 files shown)


context.ts
```diff
@@ -0,0 +1,49 @@
...
+  liveOptions = newLiveOptions;
+}
+
+export function getDeploymentId() {
+  return deploymentId;
+}
+
+export function isDenoDeploy() {
+  return _isDenoDeploy;
+}
```
lucis, line 49: I don't usually use classes, but this one might be a good fit:

- stateful
- offers an interface to get/set
- singleton

editor.tsx
```diff
@@ -34,4 +49,126 @@ export const updateComponentProps = async (
...
+    ).filter(
+      (componentPreviewData): componentPreviewData is ComponentPreview =>
+        Boolean(componentPreviewData),
+    );
+
+  const islands: ComponentPreview[] = Object.entries(manifest.islands).map(
+    mapComponentsToPreview,
+  ).filter((componentPreviewData): componentPreviewData is ComponentPreview =>
+    Boolean(componentPreviewData)
+  );
```
lucis, line 103: Out of curiosity: Why is this timed? Is there any perf bottleneck?

editor.tsx
```diff
@@ -34,4 +49,126 @@ export const updateComponentProps = async (
...
+    headers: {
+      "content-type": "application/json",
+      "Server-Timing": printTimings(),
+      ...(cache
+        ? {
+          "Cache-Control": ONE_YEAR_CACHE,
+        }
+        : {}),
+    },
+  });
```
lucis, line 120: fyi: there's `Response.json({...}, option)`

src/ComponentPreviewList.tsx
```diff
@@ -0,0 +1,78 @@
...
+export default function ComponentPreviewList(
+  { onClickComponent, registerToC }: Props,
+) {
+  const [components, setComponents] = useState<ComponentPreview[] | undefined>(
+    undefined,
+  );
+
+  useEffect(function fetchComponentAndIslands() {
+    let cancel = false;
+    if (IS_BROWSER) {
```
lucis, line 22: Do effects run on server?

src/ComponentPreviewList.tsx
```diff
@@ -0,0 +1,78 @@
+import { useEffect, useState } from "preact/hooks";
+import { asset, IS_BROWSER } from "$fresh/runtime.ts";
+import PlusIcon from "./icons/PlusIcon.tsx";
+import type { ComponentPreview } from "../editor.tsx";
+
+interface Props {
+  onClickComponent: (componentName: string) => void;
+  registerToC: (
```
lucis, line 8: I don't understand what this name means

src/ComponentPreviewList.tsx
```diff
@@ -0,0 +1,78 @@
...
+  }, []);
+
+  return (
+    <>
+      {components?.map(
+        ({ component, componentLabel, link }) => {
+          return (
+            <div
+              id={component}
+              class="mb-3 last-child:mb-0"
```
lucis, line 53: Need a `key` here?

src/EditorSidebar.tsx
```diff
@@ -1,82 +1,64 @@
...
+
   const methods = useForm<FormValues>({
-    defaultValues: mapComponentsToFormData(initialComponents),
+    defaultValues: {
+      [COMPONENTS_KEY_NAME]: mapComponentsToFormData(initialComponents),
+    },
+  });
+  const { append, fields, remove, swap } = useFieldArray({
+    control: methods.control,
+    name: COMPONENTS_KEY_NAME,
```
lucis, line 55: What do you think about moving most of this logic to a custom hook?

File might be getting too large

src/icons/PlusIcon.tsx
```diff
@@ -0,0 +1,44 @@
+import type { h } from "preact";
+
+export default function PlusIcon(
```
lucis, line 3: To the future: find a good icon library like that one TailwindUI uses:

`import { BoltIcon, ChatBubbleBottomCenterTextIcon, GlobeAltIcon, ScaleIcon } from '@heroicons/react/24/outline`

src/AddNewComponent.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+    const iframes = target.getElementsByTagName("iframe");
+    for (const iframe of iframes) {
+      observer.observe(iframe);
+    }
+
+    return () => {
+      observer.disconnect();
+    };
+    // tocComponents are the components fetched by the ComponentsPreviewList and they doesn't exist when openModal = true at the first time
```
lucis, line 62: I find it a bit hard to understand what this does. I believe it's a lazy-loading mechanism for a list of iframes that is scrollable.

Might be a good idea to comment on the function or make the function name more describing.

lucis (approved):

lucis:
Looks really good, great job!

src/ComponentPreviewList.tsx
igorbrasileiro: Nop, since this list is static.

editor.tsx
igorbrasileiro: We defined that all requests should have Timings. In future we will save these data

src/icons/PlusIcon.tsx
igorbrasileiro: Totally agree with.

src/AddNewComponent.tsx
igorbrasileiro: Yeah, totally agree. I did this to avoid fetching multiple pages, but now the pages are cached and immutable. So, this isn't a problem.

src/EditorSidebar.tsx
igorbrasileiro: Take a look at the hook, i didn't like it. The component is creating new objects unnecessarily multiple times, wdyt?

PR #12, opened 2022-09-20
feat: Improve Form fields to handle JSON Schema format

igorbrasileiro:
### Context
To be more compatible with the JSON Schema the RenderField component should handle the format, required, and type from JSON Schema.
This feature was motivated by the SEO/Metatag component schema: https://github.com/deco-pages/start/pull/6 .

### Solution
Now the RenderField component maps the JSON Schema format and checks if the property is required.

#### Changes
1. The form handles the discard, and the save button with the native submit;
2. The discard button doesn't reload the page. Now, the button resets the form state with the components value from the server.

#### Fixes
1. useEditorOperations edit the components (renamed as initialComponents) prop via componentsRef, and this may be root cause of bugs. Now the componentsRef.current has a copy of the components (initialComponents), so the components value won't be changed.

### DIFF
--- a/src/ui/JSONSchemaForm.tsx
+++ b/src/ui/JSONSchemaForm.tsx
@@ -1,6 +1,5 @@
 import type {
   JSONSchema7,
-  JSONSchema7Definition,
   JSONSchema7TypeName,
 } from "https://esm.sh/v92/@types/json-schema@7.0.11/X-YS9yZWFjdDpwcmVhY3QvY29tcGF0CmQvcHJlYWN0QDEwLjEwLjY/index.d.ts";
 import type { FunctionComponent, h } from "preact";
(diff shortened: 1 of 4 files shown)


src/ui/JSONSchemaForm.tsx
```diff
@@ -93,14 +127,24 @@ function RenderFields(
...
         return (
           <div class="flex flex-col items-start">
             <label htmlFor={fullPathField}>
               {title}
             </label>
             <Field
-              {...register(fullPathField)}
+              type={inputType}
+              pattern={pattern}
+              required={isFieldRequired}
```
lucis, line 141: Nice!

lucis (approved):

PR #13, opened 2022-09-21
feat: Parytown copy files script

igorbrasileiro:
### Context
The partytown plugin needs the lib files to instantiate the service worker/web worker, but the partytown copy files CLI doesn't work with Deno.

### Solution
Live exports createPartytownFiles script that creates partytown files at the destination folder passed from the user.
To accomplish this, the script fetches the partytown files from https://unpkg.com/browse/@builder.io/partytown@0.7.0/lib/ folder and saves them in the destination folder.

The unpkg doesn't handle list files from a folder, so, the files are hard coded for instance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ import MyTestComponent from '../components/MyTestComponent.tsx';
 ```
 ### Copy Partytown files script
 
-The partytown library need the web workers and service workers static files to work, this script copy those required files. More info: https://partytown.builder.io/copy-library-files
+The partytown library needs the web and service workers' static files to work. This script copies these required files. More info: https://partytown.builder.io/copy-library-files
 Add the `copyPartytown` task to your `deno.json` file:
 
 ```json
@@ -126,7 +126,7 @@ Then run the `copyPartytown` task with the first argument destination folder to 
 deno task copyPartytown "./static/~partytown/"
 ```
 
-To copy the debug files pass the flag `--debug`
+Pass the `--debug` flag to also copy Partytown's debug files.
 
 ```bash
 deno task copyPartytown "./static/~partytown/" -- "--debug"
--- a/README.md
+++ b/README.md
@@ -105,6 +105,33 @@ Aditionally, the import snippet will replace your clipboard content:
 ```jsx
 import MyTestComponent from '../components/MyTestComponent.tsx';
 ```
+### Copy Partytown files script
+
+The partytown library need the web workers and service workers static files to work, this script copy those required files. More info: https://partytown.builder.io/copy-library-files
+Add the `copyPartytown` task to your `deno.json` file:
+
+```json
+{
+  "tasks": {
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -105,6 +105,33 @@ Aditionally, the import snippet will replace your clipboard content:
 ```jsx
 import MyTestComponent from '../components/MyTestComponent.tsx';
 ```
+### Copy Partytown files script
+
+The partytown library need the web workers and service workers static files to work, this script copy those required files. More info: https://partytown.builder.io/copy-library-files
```
lucis, line 110:
```suggestion
The partytown library needs the web and service workers' static files to work. This script copies these required files. More info: https://partytown.builder.io/copy-library-files
```

README.md
```diff
@@ -105,6 +105,33 @@ Aditionally, the import snippet will replace your clipboard content:
...
+Add the `copyPartytown` task to your `deno.json` file:
```
lucis, line 111: Suggestion:

Move these instructions on how to add a script into a single section above, already with all the tasks that `$live` exports. Does that make sense, since now we're exporting more than one?

Also, it would make this doc only about what the script does.

Just a nitpick

scripts/copyPartytownFiles.ts
```diff
@@ -0,0 +1,70 @@
+import { ensureDirSync } from "https://deno.land/std@0.156.0/fs/mod.ts";
+import { join } from "https://deno.land/std@0.156.0/path/mod.ts";
+
+const partytownUrlPrefix = "https://unpkg.com/@builder.io/partytown@0/lib/";
```
lucis, line 4: Do you think it's worth accepting an optional `version` param here, or `partytown` doesn't really look like something that updates a lot and have mismatch features?

README.md
```diff
@@ -105,6 +105,33 @@ Aditionally, the import snippet will replace your clipboard content:
...
+}
+```
+
+Then run the `copyPartytown` task with the first argument destination folder to copy partytown files
+
+```bash
+deno task copyPartytown "./static/~partytown/"
+```
+
+To copy the debug files pass the flag `--debug`
```
lucis, line 129:
```suggestion
Pass the `--debug` flag to also copy Partytown's debug files.
```

I was understanding at first that this option helps the debug of this copy script, and not that it was related to Partytown itself. I think this makes it more clear, what do you think?

Referring to: https://developers.google.com/tech-writing/one/active-voice

lucis (approved):

scripts/copyPartytownFiles.ts
igorbrasileiro: Since our import map uses version 0.x, the script follows the same major: 0.x.

lucis: Not a big problem, just wondering. Looks good like this!

PR #14, opened 2022-09-22
feat: Handle Partytown config

igorbrasileiro:
### Context
The partytown plugin need to proxy variables from window, like gtag, dataLayer, fb, jitsu.

### Solution
Partytown plugin receives PartytownConfig parameter

#### Changes
Remove the ga plugin proxy

### DIFF
--- a/plugins/ga.ts
+++ b/plugins/ga.ts
@@ -9,7 +9,7 @@ export default function ga(options: GAOptions): Plugin {
   const main = `data:application/javascript,export default function(state){
       var ga = document.createElement("script");
       ga.type = "text/partytown";
-      ga.src = "/live/proxy/gtag/js?id=${trackingId}";
+      ga.src = "https://www.googletagmanager.com/gtag/js?id=${trackingId}";
       document.head.appendChild(ga);
 
       var snippet = document.createElement("script");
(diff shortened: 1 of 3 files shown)


plugins/ga.ts
```diff
@@ -9,7 +9,7 @@ export default function ga(options: GAOptions): Plugin {
   const main = `data:application/javascript,export default function(state){
       var ga = document.createElement("script");
       ga.type = "text/partytown";
-      ga.src = "/live/proxy/gtag/js?id=${trackingId}";
+      ga.src = "https://www.googletagmanager.com/gtag/js?id=${trackingId}";
```
lucis, line 12: Just to clarify: this was only here because @guifromrio was using Brave and it had blocked GTM, right?

lucis (approved):

plugins/ga.ts
igorbrasileiro: Yeah, probably. I had this issue, then I disabled the brave `Shield UP` and worked fined, also at chrome it works normally

guifromrio:
Isnt it best to let it working for brave users?

Sent via Superhuman ***@***.***>

igorbrasileiro:
> Isnt it best to let it working for brave users? Sent via Superhuman ***@***.***>

I don't have an opinion about it, but I think the brave is less unrepresentative.  https://gs.statcounter.com/browser-market-share

guifromrio:
OK!

guifromrio (approved):

PR #17, opened 2022-09-29
feat: Run data loaders and map component props into loaders props

igorbrasileiro:


editor.tsx
```diff
@@ -33,8 +36,51 @@ export async function updateComponentProps(
...
+
+    for (const component of components) {
+      const componentSchema =
+        LiveContext.getManifest().schemas[component.component];
+
+      if (!componentSchema) {
+        continue;
+      }
+
+      const componentRequiresLoader = Object.entries(
```
lucis, line 49:
```suggestion
      const propsThatRequireLoader = Object.entries(
```

editor.tsx
```diff
@@ -33,8 +36,51 @@ export async function updateComponentProps(
...
+      componentRequiresLoader.forEach(([property, propertySchema]) => {
+        if (typeof propertySchema !== "object") {
+          return;
+        }
+
+        if (!propertySchema?.$ref) {
+          throw new Error(`Property ${property} doesn't have $ref on schema`);
+        }
+
+        const loaderProp = component.props[property];
```
lucis, line 67: I don't understand this part. `loaderProp` is the current value for the component's prop that map to that loader?

Are you copying props from the component to the loaders, is that right?

editor.tsx
```diff
@@ -33,8 +36,51 @@ export async function updateComponentProps(
...
+        // TODO: Use uuid v4
+        const loaderInstanceName = `${propertySchema?.$ref}-${
```
lucis, line 69: From what I'm understanding here, we're regenerating the mapping of component/loaders every time we save. Is that right? Is there any data-corruption risk of doing so?

live.tsx
```diff
@@ -144,27 +156,95 @@ export function createLiveHandler<LoaderData = LivePageData>(
...
+
+      const isLoaderProp = (value: any): value is string =>
+        typeof value === "string" && value.charAt(0) === "{" &&
+        value.charAt(value.length - 1) === "}";
+
+      // map back components from database to components for the editor, merging loader props into component props
+      const editorComponents = pageData.components.map((componentData) => {
+        if (
+          Object.values(componentData.props ?? {}).some(isLoaderProp)
+        ) {
```
lucis, line 174: Do we need this `if` if you're already checking each prop in the iteration below?

live.tsx
```diff
@@ -144,27 +156,95 @@ export function createLiveHandler<LoaderData = LivePageData>(
...
+          const newComponentData = structuredClone(componentData);
+
+          for (
+            const [propName, value] of Object.entries(newComponentData.props)
+          ) {
+            if (isLoaderProp(value)) {
+              const loaderName = value.substring(1, value.length - 1);
```
lucis, line 181: Extract this logic into a function/construct that possesses all the logic regarding this. E.g:

```tsx
const loaderWrapper = {
 isLoaderProp: (value: string) => value.startsWith('{') && value.endsWith('}'),
 toPropName: (loaderName: string) => `{${loaderName}}`,
 toLoaderName: (propValue: string) => propValue.substring(1, propValue.length - 1);
}

File: live.tsx [hunk 6]
```diff
@@ -144,27 +156,95 @@ export function createLiveHandler<LoaderData = LivePageData>(
...
+              newComponentData.props[propName] = structuredClone(
+                pageData.loaders.find(({ name }) => name === loaderName) ?? {},
```
lucis, line 183: Should we also be doing some schema-stitching here? Or we're doing this later?

live.tsx
```diff
@@ -144,27 +156,95 @@ export function createLiveHandler<LoaderData = LivePageData>(
...
+          };
+        }) ?? [],
+      );
+      end("fetch-loader-data");
+
+      start("map-loader-data");
+      const loadersResponseMap = loadersResponse.reduce(
+        (result, currentResponse) => {
+          result[currentResponse.name] = currentResponse.data;
+          return result;
```
lucis, line 217: You can replace the `.map` above with `.reduce` and bundle all of this

lucis (approved):

editor.tsx
```diff
@@ -33,8 +36,51 @@ export async function updateComponentProps(
...
+          loader: propertySchema.$ref,
+          name: loaderInstanceName,
+          props: loaderProp,
+        });
+      });
+    }
+
     const res = await getSupabaseClientForUser(req).from("pages").update({
-      components: components,
+      components: { components, loaders },
```
igorbrasileiro, line 83: change this column name in the db

PR #18, opened 2022-09-30
Feature flags v0

VerasThiago:


utils/supabase.ts
```diff
@@ -0,0 +1,95 @@
+import { getSupabaseClientForUser } from "../supabase.ts";
```
lucis, line 1: TBD: What concept is this?

I think it looks like a Repository: https://learn.microsoft.com/en-us/dotnet/architecture/microservices/microservice-ddd-cqrs-patterns/infrastructure-persistence-layer-design

It'd be something like `repositories/pages.ts`. Utils is not usually a good name for stuff.

lucis (approved):

utils/supabase.ts
VerasThiago: I totally agree, I will fix this

PR #26, opened 2022-10-17
feat: New Live Layout

igorbrasileiro:
This PR implements the new Live Layout based on figma


lucis (approved):

src/EditorSidebar.tsx
```diff
@@ -95,8 +96,8 @@ export default function EditorSidebar() {
                       key={fields[index].id}
                       removeComponents={handleRemoveComponent}
                       index={index}
-                      component={component}
                       componentTitle={componentTitle}
+                      componentId={id}
```
lucis, line 100: Does the component render here before getting the `id`? Does it cause any problems?

lucis (approved):

PR #27, opened 2022-10-22
Refactor to new API

guifromrio:
This refactor simplifies live's responsibility.

It now does four things:

- Allows users to add **Editor Schemas** to their components and define **default props**;
- Renders components in **Canvas** mode for individual component coding.
- Renders dynamic pages using available components and data from the **Live Page Edge**.
- Exposes a page's metadata including props and component schemas to be used by the **Live Page Editor**.
- (Expect the addition of **Feature Flags** as our next core functionality in the near future.)

All editing functionality, thus, has been moved to `deco-pages/deco`, to be open-sourced at a later point.
This makes the surface area of the the library much smaller, allowing us to evolve the editor faster (and leveraging our own feature flags, etc).


dev.ts
```diff
@@ -69,68 +61,61 @@ export async function dev(
...
+      const schema = module?.default?.inputSchema || {};
+      return { name: filenameFromPath(loaderName), schema };
+    })
   );
+
+  return {
+    ...manifestFile,
+    loaders: loadersWithSchema,
+    components: componentsWithSchema,
+    // TODO: Islands have one type in the original manifest and other here. Double check that
```
igorbrasileiro: Here could have `schema` and the feresh manifest only has `default`

PR #31, opened 2022-11-03
Implementing dynamic routing using URLPattern

lucis:
This PR completes Live routing capabilities for dynamic routes (e.g: `/:slug/p`), without the need to manually add those routes to the project. It also organizes some functions.

It uses the very handy [URLPattern](https://developer.mozilla.org/en-US/docs/Web/API/URLPattern/exec) web API, which Fresh internally also uses it via [rutt](https://deno.land/x/rutt@0.0.13).

image

One implication we have is that now **we're fetching all production routes for every render**, which can be optimized later. I took a look at the measures and couldn't notice any difference (the timings vary a lot for `loadPage`).

**Before:**
Screen Shot 2022-11-02 at 21 18 25

**After:**
Screen Shot 2022-11-03 at 00 04 56

--

There was also some changes in `fetchPageFromId` after I realized that, even for those cases, we'd still need the page's URL to generate params because some loaders/functions will probably need it (e.g: `/blouse/p?pageId=411`, where we want to render a version of the Product Page)

### DIFF
--- a/pages.ts
+++ b/pages.ts
@@ -1,25 +1,36 @@
 import { HandlerContext } from "$fresh/server.ts";
 import { context } from "$live/server.ts";
-import { PageData } from "$live/types.ts";
+import { PageData, PageWithParams } from "$live/types.ts";
+import getSupabaseClient from "./supabase.ts";
 
 import { Page } from "./types.ts";
+import {
+  createComponent,
+  createPageForComponent,
+  exists,
+} from "./utils/component.ts";
 import { isLoaderProp, propToLoaderInstance } from "./utils/loaders.ts";
 import { path } from "./utils/path.ts";
-import { fetchPageFromComponent, fetchPageFromId, fetchPageFromPathname } from "./utils/supabase.ts";
 
-export async function loadLivePage(req: Request): Promise<Page> {
+export async function loadLivePage(
+  req: Request
+): Promise<PageWithParams | null> {
   const url = new URL(req.url);
   const pageIdParam = url.searchParams.get("pageId");
   const component = url.searchParams.get("component");
   const pageId = pageIdParam && parseInt(pageIdParam, 10);
 
-  const page = component
-    ? await fetchPageFromComponent(component)
-    : pageId
-    ? await fetchPageFromId(pageId)
-    : await fetchPageFromPathname(url.pathname, context.siteId);
+  const pageWithParams = await ((): Promise<PageWithParams | null> => {
+    if (component) {
+      return fetchPageFromComponent(component);
+    }
+    if (pageId) {
+      return fetchPageFromId(pageId);
+    }
+    return fetchPageFromPathname(url.pathname, context.siteId);
+  })();
 
-  return page;
+  return pageWithParams;
 }
 
 export async function loadData(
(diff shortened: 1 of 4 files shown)


pages.ts
```diff
@@ -83,3 +94,140 @@ export async function loadData(
...
+    .match({ site: siteId, archived: false })
+    .is("flag", null)
+    .is("archived", false);
+
+  const routes = pages?.map((page) => ({
+    page,
+    pattern: page.path,
+  })) as Array<{ pattern: string; page: Page }>;
+
+  sortRoutes(routes);
```
lucis, line 114: Allows that specific paths have priority over dynamic ones 


Screen Shot 2022-11-02 at 23 03 01

pages.ts
```diff
@@ -83,3 +94,140 @@ export async function loadData(
 
   return { ...pageData, components: componentsWithData };
 }
+
+export const fetchPageFromPathname = async (
+  path: string,
+  siteId: number
+): Promise<PageWithParams | null> => {
+  const { data: pages, error } = await getSupabaseClient()
```
tlgimenes, line 102: I liked the overall idea. However, fetching all routes is really a performance bottleneck. 

A possible trick in here is to go to the database twice, one looking for an exact match and another one filtering dynamic paths. For this to work, dinamic pages must be tagged like so. We could do this during the save process.

pages.ts
```diff
@@ -83,3 +94,140 @@ export async function loadData(
...
+};
+
+/**
+ * Sort pages by their relative routing priority, based on the parts in the
+ * route matcher
+ *
+ * Extracted from: https://github.com/denoland/fresh/blob/046fcde959041ac9cd5f2b39671c819c4af5cc24/src/server/context.ts#L683
+ */
+export function sortRoutes<T extends { pattern: string }>(routes: T[]) {
+  routes.sort((a, b) => {
```
tlgimenes, line 210: A small nit.

Create a rankRoute function that given a route, returns an integer and use this function to rank both parameters on the sort function. 

```suggestion
  routes.sort((a, b) => rankRoute(a) - rankRoute(b))

```

pages.ts
lucis: Ok, I tried implementing this and got a bit upset because Supabase's can't reuse pieces of query, and we'd have to repeat some stuff. This made me think more about this subject...

The number of production pages should be less than 50, and this is already filtered by the `flag: null, archived: false`. In that case, I suspect that the difference is minimal here, and we'd add an overhead for the dynamic pages (which is ~70% of the store).

Anyway: I'll create a separate task to investigate this problem, possibly implementing a more drastic caching strategy for production. I don't think it's worth going to the database for every page.

pages.ts
lucis: LGTY?

image

pages.ts
tlgimenes: > The number of production pages should be less than 50, and this is already filtered by the flag: null, archived: false. In that case, I suspect that the difference is minimal here, and we'd add an overhead for the dynamic pages (which is ~70% of the store).

I'm not sure about this one. We have to take into account redirects that can be possibly thousands of them. 

anyway, yeah, we can totally deal with this in the future!

pages.ts
tlgimenes: perfect!

tlgimenes (approved):
LGTM!

pages.ts
lucis: https://deco.fibery.io/Product/All-tasks-114#Tasks/20a457e0-5b86-11ed-93ab-f5261eb684bd

PR #40, opened 2022-11-09
chore: Section on html

tlgimenes:
This PR transforms the `div` section wrapper into an html 5 `section` element

### DIFF
--- a/live.tsx
+++ b/live.tsx
@@ -114,9 +114,9 @@ export function LiveSections({ sections }: PageData) {
         }
 
         return (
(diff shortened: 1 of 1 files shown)


igorbrasileiro:
LGTM, but is it right to have nested sections?

igorbrasileiro (approved):

tlgimenes:
> LGTM, but is it right to have nested sections?

According to [this](https://stackoverflow.com/questions/9916354/is-it-valid-to-have-a-section-tag-inside-another-section-tag), yes!

PR #42, opened 2022-11-09
Generating schemas from Typescript

lucis:
- Moves types (Product, ProductList) and clients from site repo to live.
- Changes `loaders` to `functions`
- Extract schema from props and put it into the manifest.


utils/denoDoc.ts
```diff
@@ -0,0 +1,363 @@
...
+const get$idFromImportUrl = (importUrl: string) => {
+  // TODO: Improve this to not include live's version when using deno.lan
+  // E.g: live@0.1.16-functions/std/commerce/types/ProductList.ts
+  return "live" + importUrl?.split("/live")[1];
+};
+
+export async function readDenoDocJson(filePath: string) {
+  const { output: rawOutput } = await exec(`deno doc ${filePath} --json`, {
+    output: OutputMode.Capture,
+  });
```
igorbrasileiro, line 39: Maybe this won't work on windows.

dev.ts
```diff
@@ -238,9 +244,69 @@ const templates = {
...
+        Boolean(inputSchema) || Boolean(outputSchema)
+    )
+    .reduce((acc, cur) => {
+      return {
+        ...acc,
+        [cur.key]: {
+          inputSchema: cur.inputSchema as JSONSchema7,
+          outputSchema: cur.outputSchema as JSONSchema7,
+        },
+      };
```
igorbrasileiro, line 308: wdyt about this suggestion?:

```suggestion
      acc[cur.key]: {
        inputSchema: cur.inputSchema as JSONSchema7,
        outputSchema: cur.outputSchema as JSONSchema7,
      };
      return acc;
```

pages.ts
```diff
@@ -171,6 +171,7 @@ export const fetchPageFromId = async (
     ? urlPattern.exec({ pathname })?.pathname.groups
     : undefined;
 
+  console.log({ matchPage, params, pathname });
```
igorbrasileiro, line 174: Remember to remove it.

std/commerce/types/Product.ts
```diff
@@ -0,0 +1,26 @@
+export interface Image {
```
igorbrasileiro, line 1: This should be here or at deco?

utils/denoDoc.ts
```diff
@@ -0,0 +1,363 @@
...
+  return denoDocOutput;
+}
+
+export function getFunctionOutputSchemaFromDocs(
+  docs: DenoDocResponse[],
+  entityName: string
+) {
+  const defaultFunctionExport = docs.find(
+    ({ name, kind }) => name === "default" && kind === "variable"
+  );
```
igorbrasileiro, line 50: This works for all this kind of exports default?
```ts
// Default exports
export default expression;
export default function functionName() { /* … */ }
export default class ClassName { /* … */ }
export default function* generatorFunctionName() { /* … */ }
export default function () { /* … */ }
export default class { /* … */ }
export default function* () { /* … */ }
```
Get these exports from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export

utils/denoDoc.ts
```diff
@@ -0,0 +1,363 @@
...
+      console.log(
+        `${entityName} doesn't export a Props interface definition, couldn't extract schema.`
+      );
+    return null;
+  }
+
+  const componentName = propsExport.location.filename
+    .split("/")
+    .pop()
+    ?.replace(/\.ts(x|)/g, "");
```
igorbrasileiro, line 131:
```suggestion
    ?.replace(/\.tsx?/g, "");
```

igorbrasileiro (commented):
Great job @lucis !

Left some commnets

igorbrasileiro:
Also, can you leave a project using this version of live?

utils/denoDoc.ts
lucis: No, we won't support every possible use case in the beginning. 

Since it's an intrinsically sensible translation, we'll have a good documentation and evolve with feedback

std/commerce/types/Product.ts
lucis: Here. 

This is the standard type definition for commerce.

I thought we had agreed that it'd be in Live, so we can import it in repos via `$live`. Do you propose to use `deco-sites/deco` instead and publish it to `deno.land`?

utils/denoDoc.ts
lucis: Added a TODO to test on Windows, good catch

utils/denoDoc.ts
lucis: Renamed this to `fileName`. This function is actually used both on `sections` and `functions`

std/commerce/types/Product.ts
igorbrasileiro: Oh, you're right, sorry.

guifromrio (approved):

PR #47, opened 2022-11-16
Add support for feature flags

guifromrio:
This enables users to register new `flags` with `data` like:

```json
{
  "id": 122,
  "name": "Test Flag",
  "state": "published",
  "data": {
    "effects": [
      {
        "key": "./functions/SelectPageEffect.ts",
        "props": {
          "pageIds": [
            289
          ]
        }
      }
    ],
    "matches": [
      {
        "key": "./functions/RandomMatch.ts",
        "props": {
          "session": false,
          "traffic": 0.5
        }
      }
    ]
  }
}
```

## How does this work

When a `flag` is in the `published` state, it's `match` functions are evaluated for each request. 
When any `match(req, ctx, props)` function is successful, this flag becomes active for this request.
Then, all `effect(req, ctx, props)` functions are executed in order, which may write to `ctx.state`. 
Finally, our `pages` rendering engine exposes the `ctx.state.selectedPageIds` API to allow effects to select pages. 

When we receive a request, we get from the database all pages matching the `path`. This may result in a list which includes both pages in `published` and in `draft` states. An effect may select a pageId, even though it is in `draft`, effectively overriding the published page.

In the future, we will remove the `published` state from the database and published pages will all be selected by a `default` flag. This way, flag management will be responsible for all page routing.

### DIFF
--- a/components/LiveAnalytics.tsx
+++ b/components/LiveAnalytics.tsx
@@ -1,9 +1,9 @@
 import { context } from "$live/live.ts";
 import Script from "https://deno.land/x/partytown@0.1.3/Script.tsx";
 import Jitsu from "https://deno.land/x/partytown@0.1.3/integrations/Jitsu.tsx";
-import type { Page } from "$live/types.ts";
+import type { Flags, Page } from "$live/types.ts";
 
-const innerHtml = ({ id, path }: Partial<Page>) => `
+const innerHtml = ({ id, path, flags }: Partial<Page> & { flags: Flags }) => `
 import { onCLS, onFID, onLCP } from "https://esm.sh/web-vitals@3.1.0";
 
 function onWebVitalsReport(event) {
@@ -16,7 +16,9 @@ function init() {
   }
 
   /* Add these trackers to all analytics sent to our server */
-  window.jitsu('set', { page_id: "${id}", page_path: "${path}", site_id: "${context.siteId}" });
+  window.jitsu('set', { page_id: "${id}", page_path: "${path}", site_id: "${context.siteId}", ${
+  Object.keys(flags).map((key) => `flag_${key}: true`).join(",")
+} });
   /* Send page-view event */
   window.jitsu('track', 'pageview');
 
@@ -33,9 +35,9 @@ if (document.readyState === 'complete') {
 };
 `;
 
-type Props = Partial<Page>;
+type Props = Partial<Page> & { flags: Flags };
 
-function LiveAnalytics({ id = -1, path = "defined_on_code" }: Props) {
+function LiveAnalytics({ id = -1, path = "defined_on_code", flags }: Props) {
   return (
     <>
       {context.isDeploy && ( // Add analytcs in production only
@@ -47,7 +49,7 @@ function LiveAnalytics({ id = -1, path = "defined_on_code" }: Props) {
 
       <Script
         type="module"
-        dangerouslySetInnerHTML={{ __html: innerHtml({ id, path }) }}
+        dangerouslySetInnerHTML={{ __html: innerHtml({ id, path, flags }) }}
       />
     </>
   );
--- a/components/LivePage.tsx
+++ b/components/LivePage.tsx
@@ -1,7 +1,7 @@
 import type { ComponentChildren } from "preact";
 import { PageProps } from "$fresh/server.ts";
 import { context } from "$live/live.ts";
-import { Page } from "$live/types.ts";
+import { LivePageData } from "$live/types.ts";
 import LiveAnalytics from "$live/components/LiveAnalytics.tsx";
 import LiveSections from "$live/components/LiveSections.tsx";
 
@@ -14,11 +14,12 @@ const EmptyPage = () => (
 );
 
 export default function LivePage({
-  data: page,
+  data,
   children,
-}: PageProps<Page | undefined> & {
+}: PageProps<LivePageData | undefined> & {
   children?: ComponentChildren;
 }) {
+  const { page, flags } = data ?? {};
   const manifest = context.manifest!;
   // TODO: Read this from context
   const LiveControls = !context.isDeploy &&
@@ -26,7 +27,7 @@ export default function LivePage({
 
   return (
     <>
-      <LiveAnalytics {...page} />
+      <LiveAnalytics {...page} flags={flags!} />
 
       {children
         ? children
@@ -39,6 +40,7 @@ export default function LivePage({
           <LiveControls
             site={{ id: context.siteId, name: context.site }}
             page={page}
+            flags={flags}
             isProduction={context.isDeploy}
           />
         )
--- a/flags.ts
+++ b/flags.ts
@@ -3,6 +3,8 @@ import { DecoManifest, Flag, LiveState } from "$live/types.ts";
 import { context } from "$live/live.ts";
 import getSupabaseClient from "$live/supabase.ts";
 import { EffectFunction, MatchFunction } from "$live/std/types.ts";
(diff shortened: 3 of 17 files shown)


std/types.ts
```diff
@@ -1,9 +1,28 @@
...
+import { MiddlewareHandlerContext } from "$fresh/server.ts";
 
 export type LoaderFunction<Props = unknown, Data = unknown, State = unknown> = (
   req: Request,
   ctx: LiveFunctionContext<State>,
   props: Props,
 ) => Promise<{ data: Data } & Partial<Pick<Response, "status" | "headers">>>;
 
-export type LoaderReturnType<O = unknown> = O;
+export type MatchDuration = "request" | "session";
```
tlgimenes, line 10: where's the LoaderReturnType?

functions/SelectPageEffect.ts
```diff
@@ -0,0 +1,20 @@
...
+  pageIds: number[];
+}
+
+const SelectPageEffect: EffectFunction<Props, LiveState> = (
+  _,
+  ctx,
+  props,
+) => {
+  if (!ctx.state.selectedPageIds) {
+    ctx.state.selectedPageIds = props.pageIds;
```
tlgimenes, line 14: Nao daria para retornar um number[] aqui para a gente evtar essa injecaco de dependencias?

live.ts
```diff
@@ -112,6 +120,10 @@ export const withLive = (
     // TODO add custom middleware optional
 
     // Prepare ctx.state for page handlers
+    ctx.state.loadFlags = async () => {
```
tlgimenes, line 123: Essas funcoes sao chamadas so no withLive, nao? se sim, a gente pode simplificar e mover essa chamada para la e tirar do ctx.state

guifromrio:
Pior que nao, elas têm que ser chamadas dentro do handler, se não elas vão
ser chamadas durante requests pra assets por ex

Sent via Superhuman iOS ***@***.***>

guifromrio:
Pode ser!

Sent via Superhuman iOS ***@***.***>

components/LivePage.tsx
```diff
@@ -34,7 +35,7 @@ export default function LivePage({
 
       {
         // Track only managed pages
-        page && <LiveAnalytics page={page} site={site} />
+        page && <LiveAnalytics page={page} flags={flags as Flags} site={site} />
```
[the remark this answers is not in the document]
igorbrasileiro: Hey, all pages will have a flag, right? So, it makes sense to me.
Otherwise, LiveAnalytics could handle flags not defined.

```diff
// at liveanatytics.tsx
- Object.keys(flags).map((key) => `flag_${key}: true`).join(",")
+ Object.keys(flags || {}).map((key) => `flag_${key}: true`).join(",") 
```

components/LiveAnalytics.tsx
```diff
@@ -1,12 +1,14 @@
...
+const innerHtml = ({ id, path }: Page, { id: siteId }: Site, flags: Flags) => `
 import { onCLS, onFID, onLCP } from "https://esm.sh/web-vitals@3.1.0";
 
 function onWebVitalsReport(event) {
   if (typeof window.jitsu === "function") {
-    window.jitsu('track', 'web-vitals', { ...event, page_id: "${id}", page_path: "${path}", site_id: "${siteId}" });
+    window.jitsu('track', 'web-vitals', { ...event, page_id: "${id}", page_path: "${path}", site_id: "${siteId}", ${
+  Object.keys(flags).map((key) => `flag_${key}: true`).join(",")
+} });
   }
```
lucis, line 12: There should be another way of injecting this info. I think flags are more "global" than web vitals and should be registered to the section. Something like [.set](https://jitsu.com/docs/sending-data/js-sdk/reference/methods-reference#set)

flags.ts
```diff
@@ -1,24 +1,80 @@
...
+  const manifest = context.manifest as DecoManifest;
 
   // TODO: Cache flags stale for 5 minutes, refresh every 30s
-  const { data: Flags, error } = await getSupabaseClient()
-    .from("flags")
+  const { data: availableFlags, error } = await getSupabaseClient()
+    .from<Flag>("flags")
     .select(
-      `id, name, audience, traffic, site!inner(name, id), pages!inner(data, path, id)`,
+      `id, name, key, state, site!inner(name, id), data`,
```
lucis, line 23: Aren't site's `name` and `id` already available at the Live Context?

I think we can avoid a DB join here.

functions/SelectPageEffect.ts
lucis: @tlgimenes onde ocorreria o "link" entre o effect da flag e a seleção da page?

live.ts
```diff
@@ -112,6 +120,10 @@ export const withLive = (
...
+      ctx.state.flags = await loadFlags(req, ctx) as Flag[];
```
lucis, line 124: Doesn't `loadFlags` already change the `ctx.state.flags`?

https://github.com/deco-cx/live/pull/47/files#diff-6a92ff4deddc2565bf7e9c8a01cb25b2083b9a782c6129d23b5faf13cd1d14a1R77

pages.ts
```diff
@@ -63,12 +82,12 @@ export async function loadLivePage(
...
   path: string,
   siteId: number,
-): Promise<PageWithParams | null> => {
+): Promise<PageWithParams[] | null> => {
   const { data: pages, error } = await getSupabaseClient()
     .from<Page & { site: number }>("pages")
     .select("id, name, data, path, state")
     .eq("site", siteId)
-    .eq("state", "published");
+    .in("state", ["published", "draft"]);
```
lucis, line 90: Why? Shouldn't we access `drafts` only when we have `?pageId=x`?

live.ts
```diff
@@ -141,13 +153,38 @@ export const withLive = (
...
+    ctx.render = (data?: Data) => {
+      const live = {
+        page: ctx.state.page,
+        flags: ctx.state.flags
+          ? ctx.state.flags.reduce((acc, flag) => {
+            acc[flag.key] = true;
+            return acc;
+          }, {} as Record<string, boolean>)
+          : {},
+      };
```
lucis, line 173: Can you extract this into more intermediate steps and giving them names?

E.g: ` const flagsByKey = ctx.state.flags?.reduce(...`

lucis (approved):
Nice, looks good!

I'd ask if we can use `ctx.state` less for this solution. It becomes harder and harder to debug the rendering process when we have a lot of mutating state.

components/LiveAnalytics.tsx
guifromrio: Ohh, my bad, I didn't notice this was a `web-vitals` specific event, d'oh. So much for programming late at night. I just stuffed it anywhere in sight 😂

live.ts
guifromrio: 😴

pages.ts
guifromrio: for now we're contemplating not having a separate `experiment` state, remember? so we don't have to have two sources of truth - being in the flag is the truth.

live.ts
guifromrio: good idea

guifromrio:
> Nice, looks good!
> 
> I'd ask if we can use `ctx.state` less for this solution. It becomes harder and harder to debug the rendering process when we have a lot of mutating state.

I have a love hate relationship with state. I'm open to not using it. But now I have to look at it differently - you're proposing we have more "functional" approach, correct?

lucis:
@guifromrio I understand the need here, and I think that a functional solution would look worse if we're using a middleware approach.

PR #50, opened 2022-11-17
feat: Web Vitals

tlgimenes:
This PR adds web-vitals report to all live sites.

It uses Google's web vitals library to add reporting to all sites. A nice thing is that it's integrated with our `pages` data, so we can cross reference web vitals to page ids and page path templates

To upgrade your site, just bump your live's version in your import_map.json

To know more: https://www.loom.com/share/f517456700124677b453745daa37dbcf


guifromrio:
Nice!! Next up is to add the active flags to this metadata!

guifromrio (approved):

PR #52, opened 2022-11-20
Major refactor to middleware approach

guifromrio:
Forgive me, rebasers, for I have sinned


live.ts
```diff
@@ -0,0 +1,118 @@
...
+  siteId: 0,
+};
+
+export const withLive = (
+  manifest: DecoManifest,
+  liveOptions?: LiveOptions,
+) => {
+  if (!liveOptions) {
+    throw new Error("liveOptions is required.");
+  }
```
lucis, line 37:
```suggestion
  liveOptions: LiveOptions,
) => {
```

lucis (approved):
The naming `withSomething` reminds/implies a high-order function that injects _Something_  but I believe it's something else here.

Suggestion: `handlePageLoad(..)`, `handleWorkbench(...)` or something like that

README.md
```diff
@@ -1,54 +1,71 @@
...
+  name: string;
+}
+
+export default function Hello({ name }: Props) {
+  return <div>Hello {name}</div>;
+}
 ```
 
-And finally, create the `live.ts` file with this:
+Finally, in order to allow fully dynamic pages, mount `live` as a handler for a catch-all route. Create `routes/[...path].tsx`:
```
lucis, line 60: Now the store will need to have `_middleware.ts` and also a catch-all route? 

This is to support the use case of only using the `middleware` and maybe using feature flags?

guifromrio: We just moved the `LiveOptions` from our own `server.ts` (that had to wrap `fresh/server.ts`) into the middleware. This way, the user doesn't have to change their main script. However, they still need to change the `dev.ts` import from fresh to live.

I agree with you regarding having to have the catch-all, it's a shame. I don't see any way to remove it right now. The good thing with this change, however, is that the handler now only has to forward `ctx.state` to the `render` function - actual fetching happens in the middleware.

guifromrio:
> The naming `withSomething` reminds/implies a high-order function that injects _Something_ but I believe it's something else here.
> 
> Suggestion: `handlePageLoad(..)`, `handleWorkbench(...)` or something like that

I actually wanted to have these sound like middleware, like you would do in express:

```
app.use(withBanana)
app.use(withOrange)
```

However we do not have such a system in fresh, and I ended up thinking it was easier for it to be bundled in one central middleware that calls everyone. I'm still open to new names.

PR #60, opened 2022-11-28
feat: Video component

tlgimenes:
This PR adds a new video component alongside it's type. 

To add a rich VideoUploadWidget to your video types, just do as you would with the image, but with the new video type instead.

```
// /sections/MySection.tsx
import { Video } from '$live/std/ui/components/Video.tsx'
import { Video as LiveVideo } from '$live/std/ui/types/Video.ts'

export interface Props {
   src: LiveVideo // This adds a VideoUploadWidget to the live editor
   alt: string
}

export default function ({ src, alt }: Props) {
   return <Video src={src} alt={alt} />
}
```

This PR depends on https://github.com/deco-sites/deco/pull/45


lucis (approved):

std/ui/components/Video.tsx
```diff
@@ -0,0 +1,34 @@
...
+    width: number;
+    height: number;
+    src: string;
+  };
+
+const Video = forwardRef<HTMLVideoElement, Props>((props, ref) => {
+  const { loading = "lazy" } = props;
+  const srcSet = getSrcSet(props.src, props.width, props.height);
+
+  return (
```
igorbrasileiro, line 22: wdyt about preloading with a link tag as fetch?

igorbrasileiro (approved):

std/ui/components/Video.tsx
tlgimenes: looks like we need to do some wild stuff: https://web.dev/fast-playback-with-preload/#preload-the-first-segment

Let's wait for the link[rel=preload][as=video] support first

PR #63, opened 2022-12-02
Enabling LiveControls in production (fixes scroll to section)

lucis:
We don't need to check if we're in production to include the `<LiveControls` island, since its code already hides the component selector:

[Código do LiveControls](https://github.com/deco-cx/live/blob/db5db5c8be254460d1c56534f265520dc3b3a2c9/components/LiveControls.tsx)
image

[Exemplo de loja com essa mudança](https://deco-sites-fashion-62dh7r5h2s60.deno.dev/)

### DIFF
--- a/components/LivePage.tsx
+++ b/components/LivePage.tsx
@@ -21,9 +21,7 @@ export default function LivePage({
 }) {
   const { page, flags } = data ?? {};
   const manifest = context.manifest!;
-  // TODO: Read this from context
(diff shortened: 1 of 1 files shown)


igorbrasileiro (approved):

guifromrio:
opa, isso devolveu o ctrl shift e?

guifromrio:
devolveu!!!

PR #70, opened 2022-12-12
Creating release task

lucis:
image

### DIFF
--- a/scripts/release.ts
+++ b/scripts/release.ts
@@ -2,7 +2,7 @@ import { exec, OutputMode } from "https://deno.land/x/exec@0.0.5/mod.ts";
 import { increment } from "https://deno.land/std@0.167.0/semver/mod.ts";
 import { Select } from "https://deno.land/x/cliffy@v0.25.5/prompt/mod.ts";
 
-await exec("git fetch");
+await exec("git fetch --tags");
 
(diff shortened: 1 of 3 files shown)


tlgimenes (approved):
AHTM. I'd just add this version to an exported variable so we can know which version of live is running. With this, we could do some "feature flags" depending on the live version

scripts/release.ts
```diff
@@ -0,0 +1,44 @@
+import { exec, OutputMode } from "https://deno.land/x/exec@0.0.5/mod.ts";
+import { increment } from "https://deno.land/std@0.167.0/semver/mod.ts";
+import { Select } from "https://deno.land/x/cliffy@v0.25.5/prompt/mod.ts";
+
+await exec("git fetch --tags");
+
+const response = await exec(`bash -c "git tag | sort -V | tail -1"`, {
+  output: OutputMode.Capture,
+});
```
igorbrasileiro, line 9: This would break on windows, wouldn't it?

igorbrasileiro (approved):

scripts/release.ts
lucis: Can you try?

igorbrasileiro: Sure!

PR #76, opened 2022-12-28
Stale-While-Revalidate cache for live functions

lucis:


utils/redisCache.ts
```diff
@@ -0,0 +1,60 @@
...
+  redis: Redis;
+  isResolved = true;
+  constructor() {
+    this.redisPromise = Promise.resolve();
+    this.redis = null as unknown as Redis;
+    this.#connect();
+  }
+
+  #connect = () => {
+    if (this.isResolved && (this.redis === null || this.redis.isClosed)) {
```
mcandeia, line 17: you might want a single flight here as you cannot guarantee a condition-free on connect

PR #79, opened 2023-01-05
chore: Bump supabase

igorbrasileiro:


supabase.ts
```diff
@@ -27,10 +26,18 @@ export default function getSupabaseClient(accessToken?: string) {
   return client;
 }
 
+/**
+ * @deprecated
```
igorbrasileiro, line 30: Should I remove this, now?

igorbrasileiro:
> We didn't use any of the removed methods? https://supabase.com/docs/reference/javascript/upgrade-guide
> 
> Also, `insert()` and `update()` alone stopped returning rows. Will that be a problem?
This depends on the projects that uses the live. If the projects uses live with supabase@1, it will maintain the same behavior, otherwise it will need to bump it. The version in the import map is only to help the editor to get metadata

PR #80, opened 2023-01-06
fix: sorting by price in deco-fashion

marialuizaaires:


std/commerce/vtex/transform.ts
```diff
@@ -46,6 +47,24 @@ export const toProductPage = (
...
+) => {
+  if (inStock(a.commertialOffer) && !inStock(b.commertialOffer)) {
+    return -1;
+  }
+
+  if (!inStock(a.commertialOffer) && inStock(b.commertialOffer)) {
+    return 1;
+  }
+
+  return a.commertialOffer.Price - b.commertialOffer.Price;
```
lucis, line 65: The command mentions `Spot Price` but I'm not sure if Spot Price is being used.

I know for a fact that `commertialOffer` has `.spotPrice` (https://github.com/deco-cx/live/blob/eeb54017ed19889e030f7db8362b639de4a66385/std/commerce/vtex/transform.ts#L154), but I'm not sure it it's the same as `.Price`. Can you double check?

lucis (approved):

std/commerce/vtex/transform.ts
marialuizaaires: Sure! I double checked and, as you suggested, I actually have to use `.spotPrice` instead of `.Price`.

PR #82, opened 2023-01-10
feat: add class that renders typographic defaults in Markdown component

marialuizaaires:


std/ui/components/Markdown.tsx
```diff
@@ -7,6 +7,9 @@ export type Props = {
 export default function Markdown({ text }: Props) {
   const body = marky(text);
   return (
-    <div class="markdown-body" dangerouslySetInnerHTML={{ __html: body }} />
+    <div
+      class="markdown-body prose"
```
igorbrasileiro, line 11: Why does this class `prose` is needed?

marialuizaaires: Porque a class `prose` estiliza o conteúdo adicionado por Markdown, necessário para os posts do blog.

Doc:  https://tailwindcss.com/docs/typography-plugin
Exemplo do que essa class faz: https://play.tailwindcss.com/uj1vGACRJA?layout=preview

@igorbrasileiro

PR #84, opened 2023-01-13
Remove domain check

guifromrio:


lucis (approved):
Needs subsequent PRs:

- In `deco` changing the contents of the `_middleware.ts` template
- In `fashion` and `start` updating live and removing domain, making sure this works fine

PR #88, opened 2023-01-25
feat: persist flags

tlgimenes:
This PR persists flags on cookies by creating one cookie per flag. 

The cookie value is the base64 of the flag object. Also, The mechanism used for invalidating the flag is by checking the last time the flag was updated. If it was updated after the cookie was set, the matcher will be rerun.

### DIFF
--- a/flags.ts
+++ b/flags.ts
@@ -7,80 +7,162 @@ import MatchRandom from "$live/functions/MatchRandom.ts";
 import MatchSite from "$live/functions/MatchSite.ts";
 import EffectSelectPage from "$live/functions/EffectSelectPage.ts";
 
+const DECO_COOKIE = "deco_flag_";
+
+export const cookies = {
+  parse: (rawCookies: string | null) => {
+    const flags: CookiedFlag[] | undefined = rawCookies
+      ?.split("; ")
+      .map((c) => c.split("="))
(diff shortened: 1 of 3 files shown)


flags.ts
```diff
@@ -7,80 +7,162 @@ import MatchRandom from "$live/functions/MatchRandom.ts";
...
 
+const DECO_COOKIE = "deco_flag_";
+
+export const cookies = {
+  parse: (rawCookies: string | null) => {
+    const flags: CookiedFlag[] | undefined = rawCookies
+      ?.split("; ")
+      .map((c) => c.split("="))
+      .filter((cookie) => cookie[0].startsWith(DECO_COOKIE))
+      .map((cookie) => JSON.parse(atob(cookie[1])));
```
lucis, line 18: Suggestion: 

```ts
import { getCookies, setCookie } from "std/http/mod.ts";
```

https://deno.land/std@0.171.0/http/cookie.ts?s=getCookies

lucis (approved):

flags.ts
lucis: I think this might not work because you're using `startsWith`, but it might be helpful for setting the cookie later

tlgimenes: yeah, I made it work now :) Thanks!

PR #95, opened 2023-02-08
Start generating own manifest

guifromrio:
First step to allow importing of manifests from other sites / libs on any deco site.


dev.ts
```diff
@@ -75,13 +78,41 @@ export default async function dev(
...
-    collectFunctions(dir),
+    collectSections(dir).then(mapWith("./sections")),
+    collectFunctions(dir).then(mapWith("./functions")),
   ]);
-  const schemas = await extractAllSchemas(sections, functions);
+
+  let schemas = await extractAllSchemas(sections, functions);
+
+  // "imports" is of format { "nameOfImport" : manifest }
+  for (const [key, importManifest] of Object.entries(imports)) {
```
igorbrasileiro, line 96: have you took a look here at the workbench API? It's a sanity check
https://github.com/deco-cx/live/blob/main/utils/workbench.ts#L21

functions/EffectSelectPage.ts
```diff
@@ -5,6 +5,10 @@ export interface Props {
   pageIds: number[];
 }
 
+export interface Return {
+  selectedPageIds: number[];
+}
+
```
igorbrasileiro, line 11: Does this interface is used?

functions/MatchDate.ts
```diff
@@ -12,11 +12,10 @@ const MatchDate: MatchFunction<Props> = (
   props,
 ) => {
   const now = new Date();
-  const start = new Date(props.start);
-  const end = new Date(props.end);
-  const isMatch = now > start && now < end;
+  const start = props.start ? now > new Date(props.start) : true;
+  const end = props.end ? now < new Date(props.end) : true;
```
igorbrasileiro, line 16: Idk if others think like me and if it's make sense, but i prefer take a look in a interval comparison like this: `lower value < higher value`. Using this for both sentences `lower value < mid value < higher value`

sections/Head.tsx
```diff
@@ -0,0 +1,85 @@
+import { asset, Head } from "$fresh/runtime.ts";
```
igorbrasileiro, line 1: Does this component here is intentional?

igorbrasileiro (approved):

functions/EffectSelectPage.ts
guifromrio: No but I would like to use it in the future! I think Candeia will solve it with the new engine

sections/Head.tsx
guifromrio: Yes

guifromrio (commented):
LGTM after changes

dev.ts
```diff
@@ -298,3 +334,10 @@ async function extractAllSchemas(
 
   return schemas;
 }
+
+// Generate live own manifest data so that other sites can import native functions and sections.
```
lucis, line 338: Noice!

lucis (approved):
@guifromrio can you write an article about this before releasing?

sections/Head.tsx
```diff
@@ -0,0 +1,85 @@
...
+  faviconUrl,
+  styleUrls,
+  inlineStyles,
+  scriptUrls,
+  inlineScripts,
+  themeColor,
+}: Props) {
+  return (
+    <Head>
+      <title>{title}</title>
```
tlgimenes, line 30: O react-helmet tem um "title template" que geramente é bem util. ref: https://github.com/nfl/react-helmet#reference-guide

tlgimenes:
LGTM, however I feel more and more we should either split the runtime from the components or merge everything into the same repo

igorbrasileiro:
> LGTM, however I feel more and more we should either split the runtime from the components or merge everything into the same repo

@guifromrio i don't have enough context, fell free to answer it.

sections/Head.tsx
guifromrio: Vc acha necessario isso melhorar nesse PR?

guifromrio:
> LGTM, however I feel more and more we should either split the runtime from the components or merge everything into the same repo

I would prefer to merge everything here

PR #96, opened 2023-02-09
fix: Routes ranking

tlgimenes:
The home page was not correctly being taking into consideration. Now, we take care of this case.


guifromrio (approved):
Solves a bug removing lines. genius

PR #98, opened 2023-02-10
chore: Prune functions

tlgimenes:
Prune functions before running so we:
1. Run repeated functions once
2. Only run functions referenced by sections

### DIFF
--- a/utils/manifest.ts
+++ b/utils/manifest.ts
@@ -135,6 +135,27 @@ export const createFunctionInstanceFromFunctionKey = (
   return functionInstance;
 };
 
+
+/**
+ * The database may have more functions than what's referenced by the sections.
+ * Maybe, this is due to a bug on the live editor's interface. Anyways, it's always
+ * good to prune and only run the necessary functions once
+ */
+const pruneFunctions = (data: PageData) => {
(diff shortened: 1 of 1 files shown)


utils/manifest.ts
```diff
@@ -135,14 +135,35 @@ export const createFunctionInstanceFromFunctionKey = (
...
+
 export async function loadPageData<Data, State extends LiveState>(
   req: Request,
   ctx: HandlerContext<Data, State>,
   pageData: PageData,
 ): Promise<PageData> {
   const { start, end } = ctx.state.t;
   const functionsResponse = await Promise.all(
-    pageData.functions?.map(async ({ key, props, uniqueId }) => {
+    pruneFunctions(pageData).map(async ({ key, props, uniqueId }) => {
```
igorbrasileiro, line 166: Wdyt about prunning sections?

igorbrasileiro (approved):

utils/manifest.ts
tlgimenes: We don't need to prune sections because the user already prunes the sections on the editor by adding/removing sections. We have to prune functions because users don't prune functions visually

PR #111, opened 2023-03-03
Ft/live resolvers

mcandeia:


engine/core/utils.ts
```diff
@@ -0,0 +1,54 @@
+export type PromiseOrValue<T> = Promise<T> | T;
+export type Entries<T> = {
+  [K in keyof T]: [K, T[K]];
+}[keyof T][];
+
+export type UnPromisify<T> = T extends Promise<infer U> ? U : T;
```
igorbrasileiro, line 6: Does this type is the same of awaited, doesn't?

PR #113, opened 2023-03-06
feat: Add header response side effect to loaders

tlgimenes:
## What's the purpose of this PR?
This PR allows loaders to return headers and status. These headers and statuses are merged following an heuristic. The only headers allowed are cache-control and vary headers. 

## How it works? 
When multiple loaders return a cache-control header on the same page, their cache-control headers must be merged. We use the more specific value first. This means that if we have two loaders returning two different cache-control headers, say `max-age=60, public, stale-while-revalidate=60` and `max-age=0, private, must-revalidate`, the final cache-control header should be `max-age=0, private, must-revalidate, stale-while-revalidate=60`

This PR adds a default cache-control to all pages: `max-age=0, s-maxage=60, public, stale-while-revalidate=3600, stale-if-error: 86400`. This value will change in case a loader returns a custom cache-control header

Also, status and vary cache headers are set. Status follows `max(loader1, loader2, 200)` logic and vary is the `join(loader1, loader2)`

### DIFF
--- a/utils/http.ts
+++ b/utils/http.ts
@@ -1,0 +1,136 @@
+export const DEFAULT_CACHE_CONTROL: CacheControl = {
+  "s-maxage": 60, // 1minute cdn cache
+  "max-age": 0, // 0s browser cache
+  "stale-while-revalidate": 3600, // 1hour
+  "stale-if-error": 24 * 3600, // 1day
+  public: true,
+};
+
+export type CacheControl = Partial<{
+  "max-age": number;
+  "s-maxage": number;
+  "stale-while-revalidate": number;
+  "stale-if-error": number;
+  "public": boolean;
+  "private": boolean;
+  "no-cache": boolean;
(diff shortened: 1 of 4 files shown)


utils/http.ts
```diff
@@ -0,0 +1,136 @@
...
+  h1: CacheControl,
+  h2: CacheControl,
+): CacheControl => {
+  const maxAge = min(h1["max-age"], h2["max-age"]);
+  const sMaxAge = min(h1["s-maxage"], h2["s-maxage"]);
+  const staleWhileRevalidate = min(
+    h1["stale-while-revalidate"],
+    h2["stale-while-revalidate"],
+  );
+  const staleIfError = min(h1["stale-if-error"], h2["stale-if-error"]);
```
lucis, line 112: Very nice!

lucis (approved):

PR #114, opened 2023-03-06
Allow two loaders of the same instance in a section.

matheusgr:
Define expected loader return type and remove $id from expect input of a section.

This will allow the same return type to be used as input in a section.


matheusgr:
Wait for deco admin update before merge ( https://github.com/deco-sites/admin/pull/215 ).

PR #117, opened 2023-03-13
[Live@v1] Add engine Core

mcandeia:
## Description

This PR adds only the core part of the new live Engine, which includes only the config resolution algorithm.

## Detailed implementation

The config resolution algorithm is based on two basic building blocks: **Resolvers** and **Resolvables**, when put together they form the **Configuration State**.

### Resolvers

Resolvers are functions that has an arbitrary name, receives a context and an input, and produces an output, resolvers can be either sync or async and can also optionally returns a resolvable. The default resolver receives an input and produces the same input as received.

```ts
export type AsyncResolver<
  T = any,
  TParent = any,
  TContext extends BaseContext = BaseContext,
> = (parent: TParent, context: TContext) => Promise<Resolvable<T> | T>;

export type SyncResolver<
  T = any,
  TParent = any,
  TContext extends BaseContext = BaseContext,
> = (parent: TParent, context: TContext) => Resolvable<T> | T;

export type Resolver<
  T = any,
  TParent = any,
  TContext extends BaseContext = BaseContext,
> = AsyncResolver<T, TParent, TContext> | SyncResolver<T, TParent, TContext>;
```

Meaning that a `Resolver<T, TInput>` is whatever function that takes TInput as a parameter and returns T or Resolvable<T> as an output.

### Resolvable

**Resolvables** are arbitrary json that contains a property named `__resolveType` which points to either: A resolver or a resolvable. When pointing to a **Resolver** the data is passed as a input to such resolver and when pointing to a **Resolvable** the algorithm starts again taking the target resolvable as input for the `__resolveType` property.

When the reference for `__resolveType` is missing, a `Dangling reference` error is thrown to sinalize that the configuration state is invalid and must be fixed.

See an example below
```json
{
	"loader-category-homens": {
		"category": "Homens",
		"__resolveType": "./loaders/VTEXProductList.ts"
	},
	"page-homens": {
		"sections": [{
			"products": {
				"__resolveType": "loader-category-homens"
			},
			"__resolveType": "./sections/ProductShelf.tsx"
		}],
		"__resolveType": "$live/pages/LivePage.tsx"
	}
}
```

Notice that the `loader-category-homens` is referenced by the `page-homens` Resolvable.

## Resolution Algorithm

The resolution algorithm takes either a Resolvable<T> or the ResolvableId as a parameter (e.g `page-homens`) and the configuration state and returns the T value to the caller, following the steps below:

1. If it is  a ResolvableId, get the Resolvable from the config state, otherwise use the provided Resolvable.
2. Check if the Resolvable has a valid __resolveType reference or it is a primitive type
3. If primitive (boolean, number, function, string), so return straight itself as a result
4. if it is a object, then for each key of the object call the resolution function recursively
5. if it is an array, then for each index call the resolution function recursively
6. when the resolver result is calculated so the resolution algorithm takes the function referenced to the __resolveType and use the resolved value as an input providing the output for the outer caller.
7. If the result of the Resolver is a Resolvable so the algorithm will be repeated until all inner objects are resolved

> In case of name clash (Resolver and Resolvable with the same name) a Resolver function takes precedence.

Important:
If the target object that needs to be resolved does not contains the `__resolveType` property so the resolver algorithm consider as a resolved object. Which means that if you have inner objects that needs to be resolved in the same resolution cycle, so you need to manually reference the `resolve` function which is the default resolver. We chose to do that to avoid unnecessary computation (going too far when there's nothing to resolve) when possible (and there's a ambiguity between resolved objects and resolvable-eligible objects).

The following object does not resolve the inner `sections` array
```json
{
    "name":  "my-config-name",
    "sections": [{ "__resolveType": "section#123"}]
}
```

To fix that use the `resolve` func.

```json
{
    "name":  "my-config-name",
    "sections": [{ "__resolveType": "section#123"}],
   "__resolveType": "resolve"
}
```


lucis:
The documentation is very clear. Don't forget to save that somewhere (README.md)?

import_map.json
```diff
@@ -4,6 +4,7 @@
     "std/": "https://deno.land/std@0.170.0/",
     "$fresh/": "https://deno.land/x/fresh@1.1.2/",
     "preact-render-to-string": "https://esm.sh/*preact-render-to-string@5.2.4?pin=102",
+    "json-schema": "https://esm.sh/@types/json-schema@7.0.11?pin=102",
```
lucis, line 7: I think we started using `deps.ts` file instead of adding it to the manifest. IIRC, this will require the site to include the `json-schema` entry in the manifest

engine/core/resolver.ts
```diff
@@ -0,0 +1,145 @@
...
+>;
+
+async function object<T>(
+  obj: ResolvableObj<T> | T,
+  resolve: <K extends keyof T>(
+    resolvable: Resolvable<T[K]>,
+  ) => PromiseOrValue<T[K]>,
+): Promise<T> {
+  if (obj instanceof Date) {
+    return obj;
```
lucis, line 74: Should we expect that resolving a config will return a serializable object?

lucis: In that case, this should probably return a number or string

lucis (approved):
Looks good!

I can understand what the code does, but I don't have ground knowledge to suggest any improvements.

One thing that comes to mind is: what do you think about adding a `verbose` or `debug` flag that outputs all steps performed in the resolution? If ~~god forbid~~ I'd have to fix a bug here, the first thing I'd do would be set up something like that to understand what's going on.

Amazing job, @mcandeia! It's a new milestone for deco.cx

engine/core/resolver.ts
mcandeia: Agree, but seems that date is serializable to ISO8601 format by default, so we are fine, right?

image

import_map.json
mcandeia: >One thing that comes to mind is: what do you think about adding a verbose or debug flag that outputs all steps performed in the resolution? If god forbid I'd have to fix a bug here, the first thing I'd do would be set up something like that to understand what's going on.

Definitely! I think we can add this in the upcoming PRs, WDYT?

> I think we started using deps.ts file instead of adding it to the manifest. IIRC, this will require the site to include the json-schema entry in the manifest

Will do, thanks!

engine/core/README.md
igorbrasileiro, line 1: Wdyt about adding a link at /Readme.md to this file?

mcandeia: I think it is something tied to "contributors" more than  "end-users", I'd rather prefer keep the root README.md focused on the end-users, does this make sense?

engine/core/README.md
```diff
@@ -0,0 +1,95 @@
...
+> = AsyncResolver<T, TParent, TContext> | SyncResolver<T, TParent, TContext>;
+```
+
+Meaning that a `Resolver<T, TInput>` is whatever function that takes TInput as a parameter and returns T or Resolvable<T> as an output.
+
+### Resolvable
+
+**Resolvables** are arbitrary json that contains a property named `__resolveType` which points to either: A resolver or a resolvable. When pointing to a **Resolver** the data is passed as a input to such resolver and when pointing to a **Resolvable** the algorithm starts again taking the target resolvable as input for the `__resolveType` property.
+
+When the reference for `__resolveType` is missing, a `Dangling reference` error is thrown to sinalize that the configuration state is invalid and must be fixed.
```
igorbrasileiro, line 35: Wdyt about adding an example with this dangling reference? You can move this to after the example below and add an example

engine/core/resolver.ts
lucis: Yeah, that will work

The only reason I brought it up is that Fresh itself has a limitation of which type of data can be passed as props to an Island, for example: https://fresh.deno.dev/docs/concepts/islands I

igorbrasileiro (approved):
LGTM.

I saw some files that contains the same code, but formatted, is it ok?

engine/core/README.md
mcandeia: Done

engine/core/README.md
igorbrasileiro: Oh, great.

PR #121, opened 2023-03-14
[Live@v1] All block implementations

mcandeia:


flags/audience.ts
```diff
@@ -0,0 +1,20 @@
+import { Flag } from "$live/blocks/flag.ts";
+import { Matcher } from "$live/blocks/matcher.ts";
+import { Resolvable } from "$live/engine/core/resolver.ts";
+import { Handler } from "$live/blocks/handler.ts";
+
+export interface Audience {
+  matcher: Matcher;
+  name: string;
```
lucis, line 8: Why `name` here? Isn't this the `alias` metadata prop we've talked before?

lucis: I've seen the "Everyone" below.

Isn't the fileName `everyone.ts` that would give name to this block?

pages/LivePage.tsx
```diff
@@ -0,0 +1,25 @@
+import { Section } from "$live/blocks/section.ts";
+import LiveAnalytics from "$live/components/LiveAnalytics.tsx";
+
+export interface Props {
+  sections: Section[];
+}
+
+// FIXME @author Marcos V. Candeia missing live controls
```
lucis, line 8: @guifromrio will probably want to inline the script here instead

lucis (approved):

flags/audience.ts
mcandeia: I think it is different perspectives. The `alias` we talked about is for the `business user`, the developer doesn't have access to such alias. This name is the Flag name, same for the audience. An audience has a name that identifies it uniquely.

mcandeia: > I've seen the "Everyone" below.
> Isn't the fileName everyone.ts that would give name to this block?

Yes. Think about Java classes. The audience is a classe that takes an audience config as a parameter. The everyone is an audience (extends Audience) but provides some parameters as default value, which is the case for the name (Everyone) and the matcher (Always True).

But everyone has a different "constructor".

mcandeia: Notice that Audience is a class that extends Flag and so on and so fourth

PR #135, opened 2023-03-25
V0 Compatibility layer

mcandeia:
This pr adds a few endpoints and redirects to fulfill the editor data api from livev0


blocks/function.ts
```diff
@@ -37,7 +37,7 @@ const functionBlock: BlockForModule<FunctionModule> = {
       $live: TConfig,
       ctx: HttpContext<any, any, HandlerContext<any, TState>>,
     ) => {
-      const global = await ctx.resolve({ __resolveType: "globals" });
+      const global = await ctx.resolve({ __resolveType: "accounts" });
```
mcandeia, line 40: changed from "globals" to "accounts" to match the real semantics

blocks/route.ts
```diff
@@ -81,20 +81,15 @@ const mapHandlers = (
...
         const $live = await ctxResolver(key);
-        return val!(request, {
-          ...context,
-          state: {
-            ...context.state,
-            $live,
-            resolve: ctxResolver,
-          },
-        });
+        context.state = { ...context.state, $live, resolve: ctxResolver };
```
mcandeia, line 84: I had to set the context.state so that I could modify the state for the downstream resolvers. Previously It wasn't possible because I was copying the reference.

compatibility/v0/editorData.ts
```diff
@@ -0,0 +1,611 @@
+import { Schemas } from "$live/engine/schema/builder.ts";
```
mcandeia, line 1: this file is supposedly used only for compatibility with the current admin version.

It is a bunch of functions copied from the live 0.x that serve as a "polyfill" for the admin as it is currently

components/LiveControls.tsx
```diff
@@ -86,8 +86,11 @@ const main = () => {
 
     switch (data.type) {
       case "scrollToComponent": {
-        document
-          .getElementById(data.args.id)
+        (document
+            .getElementById(data.args.id) ?? data.args.alternateId
+          ? document
+            .getElementById(data.args.alternateId)
+          : undefined)
```
mcandeia, line 93: I've changed that because the concept of `uniqueId` does not exists in the new engine. So I'm trying to support both behaviors providing an alternateId that is used as a section index

dev.ts
```diff
@@ -110,6 +110,7 @@ export default async function dev(
     onListen?: () => void;
   } = {},
 ) {
+  const site = siteId ?? -1;
```
mcandeia, line 113: same as defining as not defined

engine/configstore/supabase.ts
```diff
@@ -11,14 +11,14 @@ function sleep(ms: number) {
 const sleepBetweenRetriesMS = 100;
 const refetchIntervalMSDeploy = 5_000;
 
-const fetchConfigs = (site: number) => {
-  return getSupabaseClient().from("configs").select("config").eq(
+const fetchConfigs = (site: string) => {
+  return getSupabaseClient().from("configs").select("state").eq(
```
mcandeia, line 15: We have changed this from "config" to "state"

engine/configstore/supabaseLegacy.ts
```diff
@@ -5,48 +5,69 @@ import { Resolvable } from "$live/engine/core/resolver.ts";
...
+  if (key.includes("Global")) {
     return {
-      ...newProps,
-      __resolveType: key
-        .replace("./", `${ns}/`),
+      page: {
+        __resolveType: globalSections[key],
+      },
+      __resolveType: "$live/sections/PageInclude.tsx",
     };
```
mcandeia, line 49: Sections that used to be "Global" now are converted to a section named "PageInclude" which just render the inner sections on it.

engine/configstore/supabaseLegacy.ts
```diff
@@ -67,67 +88,83 @@ interface AudienceFlag {
...
     "deco-sites/std/accounts/vtex.ts",
   "deco-sites/std/sections/configOCC.global.tsx":
     "deco-sites/std/accounts/occ.ts",
   "deco-sites/std/sections/configShopify.global.tsx":
     "deco-sites/std/accounts/shopify.ts",
+  "deco-sites/std/sections/configVNDA.global.tsx":
+    "deco-sites/std/accounts/vnda.ts",
+  "deco-sites/std/sections/configYourViews.global.tsx":
+    "deco-sites/std/accounts/yourViews.ts",
+};
```
mcandeia, line 102: these are the wellKnown global sections that is supposed to be accounts instead.

engine/configstore/supabaseLegacy.ts
```diff
@@ -238,6 +293,73 @@ export const newSupabaseProviderLegacyDeploy = (
   };
 };
 
+const matchesToMatchMulti = (matches: Flag["data"]["matches"], ns: string) => ({
+  op: "or",
+  matchers: matches.map((m) => ({
+    ...m.props,
+    __resolveType: includeNamespace(m.key, ns).replace("functions", "matchers"),
```
mcandeia, line 300: matchers used to be within the functions folder. so we need to rename

engine/configstore/supabaseLegacy.ts
```diff
@@ -238,6 +293,73 @@ export const newSupabaseProviderLegacyDeploy = (
...
+  })),
+  __resolveType: "$live/matchers/MatchMulti.ts",
+});
+
+const flagsToConfig = (
```
mcandeia, line 305: flags should be mapped to audiences

engine/configstore/supabaseLegacy.ts
```diff
@@ -238,6 +293,73 @@ export const newSupabaseProviderLegacyDeploy = (
...
+    };
+  }, entrypoint);
+};
+const pagesToConfig = (
+  p: Page[],
+  flags: Pick<Flag, "key" | "data" | "name">[],
+  ns: string,
+) => {
+  const configs = p.sort((pageA, pageB) =>
+    pageA.state === "global" ? -1 : pageB.state === "global" ? 1 : 0
```
mcandeia, line 357: global pages should be mapped first to be available in other components that depends on it

engine/fresh/manifestGen.ts
```diff
@@ -47,6 +47,18 @@ export const defaultRoutes: {
...
+    block: "routes",
+    from: "$live/routes/live/workbench.ts",
+    key: "./routes/live/workbench.ts",
+    ref: "$live_workbench",
+  },
+  {
+    block: "routes",
+    from: "$live/routes/live/editorData.ts",
+    key: "./routes/live/editorData.ts",
+    ref: "$live_editorData",
```
mcandeia, line 60: workbench and editor data are now added automatically

engine/schema/builder.ts
```diff
@@ -180,12 +180,15 @@ export const newSchemaBuilder = (initial: SchemaData): SchemaBuilder => {
       const canonical = canonicalFileWith(base, namespace);
```
mcandeia, line 180: the change in this file was necessary to avoid typings being created to primitive values like `null` or `string`. Previously it was added as a `definition` and referenced in the other components which makes no sense since it isn't reused across them.

engine/schema/transform.ts
```diff
@@ -31,6 +31,7 @@ export const inlineOrSchemeable = async (
 };
```
mcandeia, line 31: the jsDoc schema for type alias was not being used. so it was a bug fix

handlers/fresh.ts
```diff
@@ -12,8 +12,13 @@ export const isFreshCtx = <TState>(
   return typeof (ctx as HandlerContext).render === "function";
 };
 export default function Fresh(page: FreshConfig) {
-  return (_: Request, ctx: ConnInfo) =>
-    isFreshCtx(ctx)
+  return (req: Request, ctx: ConnInfo) => {
+    const url = new URL(req.url);
+    if (url.searchParams.get("asJson") !== null) {
+      return Response.json(page);
+    }
```
mcandeia, line 19: if `?asJson` is provided, so instead of rendering the component a JSON object will be returned

handlers/routesSelection.ts
```diff
@@ -7,8 +7,7 @@ import { CookiedFlag, cookies } from "$live/flags.ts";
 import { Audience } from "$live/flags/audience.ts";
```
mcandeia, line 7: the force audience was removed in favor of preview page by id

handlers/routesSelection.ts
```diff
@@ -138,17 +100,15 @@ export default function RoutesSelection({ flags }: SelectionConfig): Handler {
...
-    });
-    const [routes, overrides] = (pageAudience ? [pageAudience] : audiences)
+    const [routes, overrides] = audiences
       .reduce(
         ([routes, overrides], audience) => {
           // check if the audience matches with the given context considering the `isMatch` provided by the cookies.
           const isMatch = audience.matcher({
             ...matchCtx,
-            isMatchFromCookie: flags.get(audience.name)?.isMatch,
+            isMatchFromCookie: isNoCache
```
mcandeia, line 109: allowing CTRL+SHIFT+R to not consider cookie values

handlers/routesSelection.ts
```diff
@@ -199,7 +159,7 @@ export default function RoutesSelection({ flags }: SelectionConfig): Handler {
     const resp = await server(req, connInfo);
 
     // set cookie for the flags that has changed.
-    if (flagsThatShouldBeCookied.length > 0) {
+    if (flagsThatShouldBeCookied.length > 0 && resp.status < 300) { // errors and redirects have immutable headers
```
mcandeia, line 162: for some reason errors and redirects have immutable headers, so we can't set them

Source: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Basic_concepts#guard

Notice that's not all errors but the Response.errors() and Response.redirect was. I've tested it manually setting the redirect status code which results in the same error, so I think it is internally set to immutable

pages/LivePage.tsx
```diff
@@ -1,24 +1,29 @@
...
-import { usePageContext } from "$live/routes/[...catchall].tsx";
+import { context } from '$live/live.ts';
+import { usePageContext } from '$live/routes/[...catchall].tsx';
+import LiveControls from "../components/LiveControls.tsx";
 
 export interface Props {
   sections: Section[];
 }
 
+export function renderSection(
```
mcandeia, line 11: exported to be used on PageInclude.tsx

routes/_middleware.ts
```diff
@@ -15,6 +29,36 @@ export const handler = async (
...
+  if (
+    !url.pathname.startsWith("/live/previews") && url.searchParams.has("key") &&
+    !url.searchParams.has("editorData")
+  ) {
+    return previewSection(url, url.searchParams.get("key")!);
+  }
+  if (
+    !url.pathname.startsWith("/live/previews") &&
+    url.searchParams.has("pageId") &&
+    !url.searchParams.has("editorData")
```
mcandeia, line 58: redirects to editor data and preview page depending on the specified query params

routes/live/previews/[...block].tsx
```diff
@@ -19,13 +34,23 @@ const buildObj = (
...
+    partial[arrKey] ??= [];
+    const idxNumber = +idx.substring(0, idx.length - 1);
+    rest.unshift(idxNumber.toString());
+    key = arrKey;
+  } else {
+    partial[first] ??= {};
+  }
   if (rest.length === 0) {
-    partial[first] = isNumber.test(value) ? +value : value;
+    partial[key] = isNumber.test(value) ? +value : value;
```
Comment (mcandeia) [inline routes/live/previews/[...block].tsx:48]: supporting array keys `banner[0].x=10`

compatibility/v0/editorData.ts
```diff
@@ -0,0 +1,382 @@
...
+      ...def,
+      anyOf: def.anyOf.map((obj) => flat(obj as JSONSchema7, schema, memo)),
+    };
+  }
+  if (def.type === "array") {
+    return { ...def, items: flat(def.items as JSONSchema7, schema, memo) };
+  }
+  const props: Record<string, JSONSchema7> = {};
+  for (const [propName, propValue] of Object.entries(def?.properties ?? {})) {
+    const flatObj = flat(propValue as JSONSchema7, schema, memo);
```
lucis, line 123: 👏  for thie file

components/LiveControls.tsx
lucis: Suggestion: Extract these conditions into consts.

engine/configstore/supabaseLegacy.ts
```diff
@@ -5,51 +5,73 @@ import { Resolvable } from "$live/engine/core/resolver.ts";
...
+  }
+  if (key.includes("Global")) {
     return {
-      ...newProps,
-      __resolveType: key
-        .replace("./", `${ns}/`),
+      page: {
+        __resolveType: globalSections[key],
+      },
+      __resolveType: "$live/sections/PageInclude.tsx",
```
lucis, line 48: Nice

engine/configstore/supabaseLegacy.ts
```diff
@@ -67,67 +89,88 @@ interface AudienceFlag {
...
   "deco-sites/std/sections/configVTEX.global.tsx":
     "deco-sites/std/accounts/vtex.ts",
   "deco-sites/std/sections/configOCC.global.tsx":
     "deco-sites/std/accounts/occ.ts",
   "deco-sites/std/sections/configShopify.global.tsx":
     "deco-sites/std/accounts/shopify.ts",
+  "deco-sites/std/sections/configVNDA.global.tsx":
+    "deco-sites/std/accounts/vnda.ts",
+  "deco-sites/std/sections/configYourViews.global.tsx":
+    "deco-sites/std/accounts/yourViews.ts",
```
lucis, line 102: What happens when we have a global config that's not here?

engine/fresh/manifest.ts
```diff
@@ -175,37 +166,50 @@ export const $live = <T extends DecoManifest>(m: T): T => {
...
+        integrityCheck(resolver.getResolvers(), resolvables);
+      },
+    );
+  }
   // should be set first
   context.configResolver = resolver;
   context.manifest = newManifest;
+  console.log(
+    `Starting live: siteId=${context.siteId} site=${context.site}`,
+  );
```
lucis, line 212: If we are to ditch siteId (as number), could we be more "depreacted-ly" here?

lucis (approved):

engine/configstore/supabaseLegacy.ts
mcandeia: needs to be added or it will not work, functions will start to receive `undefined`

mcandeia: I can create a stub value that just receive and returns, should be enough

engine/fresh/manifest.ts
mcandeia: yes, we can! I think it should be addressed in the next steps when blocks UI was released

PR #137, opened 2023-03-29
v1 Migration Script

mcandeia:
Como testar

1. clone deco-sites/fashion
2. vá até o diretório clonado
3. rode `deno run -A https://denopkg.com/deco-cx/live@fc11cc12d5a6d2870758683559dce92a0ff4363f/scripts/upgrade.ts`

É possível ter erros relacionados a versão do std pois ela não foi lançada ainda.

### DIFF
--- a/components/LiveControls.tsx
+++ b/components/LiveControls.tsx
@@ -97,12 +97,15 @@ const main = () => {
 
     switch (data.type) {
       case "scrollToComponent": {
-        (document
-            .getElementById(data.args.id) ?? data.args.alternateId
+        const findById = document
+          .getElementById(data.args.id);
+
+        const findByAlternateId = data.args.alternateId
           ? document
             .getElementById(data.args.alternateId)
-          : undefined)
-          ?.scrollIntoView({ behavior: "smooth" });
+          : undefined;
+
+        (findById ?? findByAlternateId)?.scrollIntoView({ behavior: "smooth" });
 
         return;
       }


meta.json
```diff
@@ -1,3 +1,3 @@
 {
-  "version": "1.0.0-rc.1"
+  "version": "01ac779045fc1f81122e7a544817cf007c5e2528"
```
lucis, line 2: Aware?

scripts/upgrade.ts
```diff
@@ -0,0 +1,254 @@
...
+  const devTs = join(Deno.cwd(), "dev.ts");
+  if (!(await exists(devTs))) {
+    throw new UpgradeError(`${devTs} not found`);
+  }
+  const devTsContent = await Deno.readTextFile(devTs);
+  const updateFreshGen = devTsContent.replaceAll(
+    "fresh.gen.ts",
+    "live.gen.ts",
+  );
+  const id = await siteId();
```
lucis, line 121: Really cool!

WDYT about placing `v0-to-v1.ts` in another file and having also a simple way to declare future upgrades (that might only be a version bump)?

scripts/upgrade.ts
```diff
@@ -0,0 +1,254 @@
...
+
+  return {
+    from: {
+      path: mainTs,
+      content: mainTsContent,
+    },
+    to: {
+      path: mainTs,
+      content: mainTsContent.replace(
+        `import manifest from "./fresh.gen.ts";\n`,
```
lucis, line 156: FYI: Might not work on Windows projects, not sure

lucis (approved):

scripts/upgrade.ts
mcandeia: I think it will because it is how deno interprets imports, should not be related to file system separators

scripts/upgrade.ts
mcandeia: the way it is implemented already support to split in many files but I didn't do that for now, can we defer a bit this implementation?

PR #139, opened 2023-04-02
Support workbench GET and /live/_meta endpoint

mcandeia:


scripts/upgrade.ts
```diff
@@ -175,7 +175,7 @@ const v1: UpgradeOption = {
       .all([
         updateImportMap(
           meta.version,
-          "1.0.0-rc.0",
+          "1.0.0-rc.5",
```
igorbrasileiro, line 178: Does this value is right, rc.5?

igorbrasileiro (approved):

scripts/upgrade.ts
mcandeia: yes, the std version

PR #143, opened 2023-04-05
[ft] Inline loadData on sections

mcandeia:
Improve JSONSchema support (because of loadData requires it)

- Add support for `Omit` and `Pick
- Add support for `Intersection` types (A&B)
- Fixes the recursion break


blocks/section.ts
```diff
@@ -1,18 +1,166 @@
...
+  metadata: {
+    component: resolver,
+    resolveChain,
+    id: resolveChain.length > 0 ? resolveChain[0] : undefined,
+  },
 });
 
+const sectionBlock: Block<SectionModule> = {
+  type: "sections",
+  introspect: async (
```
mcandeia, line 63: what this does is:

Get the props from the default function and the loadData function.

Create a JSONSchema which is the intersection of both props except all keys from the loadData return

blocks/section.ts
```diff
@@ -1,18 +1,166 @@
...
+      >,
+      HttpContext
+    > => {
+    const componentFunc = componentWith(resolver, mod.default);
+    const loadDataFunc = mod.loadData;
+    if (!loadDataFunc) {
+      return (
+        props: TProps,
+        { resolveChain }: BaseContext,
+      ): PreactComponent<any, TProps> => componentFunc(props, resolveChain);
```
mcandeia, line 134: if does not contains loadDataFunc keep the old behavior

PR #145, opened 2023-04-05
Fix namespace on git remote from https.

matheusgr:


lucis (approved):

utils/namespace.ts
```diff
@@ -7,9 +7,18 @@ const namespaceFromGit = async (): Promise<string | undefined> => {
...
+    // HTTPS
+    // Fetch URL: https://github.com/deco-sites/example
+    regex = /.*https:\/\/github\.com\/(.+)/;
+  }
+
+  const result = fetchUrlLine.match(regex);
+  if (!result || (result.length < 1)) {
+    return undefined;
+  }
+  return "/" + result[1].trimEnd();
```
mcandeia, line 21: Not sure but I think the namespace should not start with `/` it should be something like `deco-sites/fashion`

mcandeia (changes requested):

utils/namespace.ts
matheusgr: I kept the current implementation, but it breaks import_map as is.

mcandeia (approved):

mcandeia:
@matheusgr reverti pq quando fuit estar no mac meu import_map botou um undefined la
image


A gente ve junto hoje!

mcandeia:
image
essa é a saida do meu comando

PR #149, opened 2023-04-06
Try a aggressive regex to match org/repo.

matheusgr:


matheusgr:
Current implementation works if you clone using recommended github SSH/HTTPS command. That should be enough for now.

PR #153, opened 2023-04-09
Minor fixes before v1 and Section Loader

mcandeia:
- A lot of minor fixes regarding block typings
- Implementing some missing transformations on JSONSchema
- Fixes overfetching on middleware (the middleware is querying database when running in localhost even when JS or static assets are being fetched)
- Remove supabase local implementation in favor of refreshing config once per request


blocks/route.ts
```diff
@@ -56,6 +54,47 @@ const isConfigurableRoute = (
...
+    request: Request,
+    context: MiddlewareHandlerContext<LiveConfig<any, LiveState>>,
+  ) {
+    const { start, end, printTimings } = createServerTimings();
+    context.state.t = { start, end, printTimings };
+    const url = new URL(request.url);
+    if (
+      url.pathname.startsWith("/_frsh") ||
+      url.pathname.startsWith("~partytown") ||
+      url.searchParams.has("__frsh_c")
```
lucis, line 71: Suggestion: Extract into a const with a name that explains why this is needed

lucis (approved):

PR #176, opened 2023-04-19
Fix titles and schema references on windows

mcandeia:


guitavano:
Tested at https://github.com/deco-sites/starting/pull/102 using windows.

PR #183, opened 2023-04-26
feat: LivePage Preview Editor

igorbrasileiro:
This PR adds Editor Preview for LivePage and this Preview sends events to be handled via Post Message.

Also, this PRs fixes the preview override props


pages/LivePage.tsx
```diff
@@ -14,21 +14,38 @@ import { JSX } from "preact/jsx-runtime";
...
   return (
+    // data-manifest-key used at preview
     <section
       id={`${metadata?.component}-${idx}`}
       data-manifest-key={metadata?.component}
     >
+      {controls}
       <Section {...props} />
     </section>
   );
```
mcandeia, line 51: since you're receiving all the properties you need for render the controls `<SectionControls>` in this function how about partial apply this function with the `preview` boolean and then create an empty element when preview is false? e.g

```ts
const renderSectionFor = (preview: boolean = false) => {
  const Controls = preview ? SectionControls : () => null;
  return (
    { Component: Section, props, metadata }: Props["sections"][0],
    idx: number,
  ) => {
    return (
      <section
        id={`${metadata?.component}-${idx}`}
        data-manifest-key={metadata?.component}
      >
        <Controls {...metadata} index={idx} />
        <Section {...props} />
      </section>
    );
  };
};
```

Then you can just use it directly `sections.map(renderSectionFor(preview))`

components/LivePagePreview.tsx
```diff
@@ -0,0 +1,462 @@
...
+          })}
+        >
+          <PreviewIcon id="chevron-down" />
+        </button>
+        {
+          /*
+            <button>
+              <PreviewIcon id="components" />
+            </button>
+            */
```
mcandeia, line 310: remove?

components/LivePagePreview.tsx
```diff
@@ -0,0 +1,462 @@
...
+        /*
+        section[data-manifest-key]:hover {
+          position: relative;
+          outline-style: solid;
+          outline-color: #2E6ED9;
+          outline-width: 2px;
+          margin-left: 2px;
+          margin-right: 2px;
+        }
+      */
```
mcandeia, line 348: remove?

mcandeia (changes requested):

pages/LivePage.tsx
mcandeia: and you can also export the default renderSection 

export const renderSection = renderSectionFor();

components/LivePagePreview.tsx
igorbrasileiro: Fixed

pages/LivePage.tsx
igorbrasileiro: Fixed

mcandeia (approved):

igorbrasileiro:
@mcandeia does it make sense for the event to send the section's path?

For instance, is only possible to edit sections, but in the future, we would edit sections from a slot and the path could be `{slotname}_sections`. Does it make sense?

lucis (approved):

components/LivePagePreview.tsx
igorbrasileiro, line 1: LivePageEditor.tsx

mcandeia (approved):

PR #195, opened 2023-05-02
Use site.json to get namespace

mcandeia:
# Description

This pr fixes the problem of using the ${org}/${user} to define the store's namespace. This is a problem since contributors will start the project from theirs own fork.

With these changes the namespace is always read from the `site.json`


dev.ts
```diff
@@ -183,25 +163,28 @@ export default async function dev(
...
     },
   });
-  await raw.pipeTo(proc.stdin.writable);
+
+  await raw.pipeTo(proc.stdin);
   const out = await proc.output();
-  await proc.status();
-  proc.close();
+  await proc.status;
+  proc.kill();
```
igorbrasileiro, line 185: Ensure that this process will close when parent close.

igorbrasileiro (approved):

scripts/upgrade.ts
```diff
@@ -5,10 +5,35 @@ import {
...
+  if (fetchUrlLine.startsWith("http")) { // http clone
+    const fetchUrl = new URL(fetchUrlLine);
+    return fetchUrl.pathname.substring(1).replace(".git", "").trimEnd(); // remove .git
+  }
+  if (fetchUrlLine.startsWith("git")) {
+    const [_ignoreGitUrl, nsAndGit] = fetchUrlLine.split(":");
+    const [namespace] = nsAndGit.split(".");
+    return namespace.trimEnd();
+  }
+  return fetchUrlLine.replace(":", "/").trimEnd();
```
lucis, line 34: What's this case?

lucis (approved):

PR #198, opened 2023-05-03
Remove duplicated preview on resolveChain

mcandeia:


pages/LivePage.tsx
```diff
@@ -49,6 +49,7 @@ interface UseSlotSection {
   used: boolean;
 }
 
+const useSlotSection = "$live/sections/UseSlot.tsx" as const;
```
igorbrasileiro, line 52: "UseSlotResolveType" wdyt about this name?

igorbrasileiro (commented):
LGTM Overall, i will test resolver chain at inline editor

pages/LivePage.tsx
mcandeia: Changed to `USE_SLOT_SECTION_KEY` sounds better?

PR #210, opened 2023-05-07
Add autoupdate script

mcandeia:
# Description

This pr uses an opensource package that checks package versions in a given deno registry.

I'm using it to check whenever `$live` and `deco-sites/std` got updated and then prompt the user asking if they would like to upgrade to the latest version.


lucis (approved):

utils/update.ts
```diff
@@ -0,0 +1,66 @@
+export * from "https://denopkg.com/hayd/deno-udd@master/mod.ts";
```
igorbrasileiro, line 1: Where this package is being used?

mcandeia: Didn’t notice it was there, thanks, will remove!

PR #234, opened 2023-05-16
fix: page.id

igorbrasileiro:
This fixes the page metadata.id, allowing to open the admin by pressing `.`


handlers/devPage.ts
```diff
@@ -17,18 +17,19 @@ export default function DevPage(devConfig: DevConfig) {
...
       context.isDeploy
     ) {
       if (!referer || !isOnAdmin) {
-        if (!pageId) {
+        if (!pageParent) {
           return Response.error();
         }
         // redirect
         return Response.redirect(
-          adminUrlFor(+pageId, context.siteId),
```
igorbrasileiro, line 31: did get it why need to increment pageId, since it's a string.

mcandeia (approved):

PR #254, opened 2023-05-28
Add asRaw transform and start library on LivePageEditor

guifromrio:
This PR adds an `asRaw` query string to html that returns only the first child element of <body>. This allows rendering a safe, server-side HTML only version of blocks previews. 

This also adds an initial visual library UI which uses the asRaw rendering to render multiple blocks. Ideally, blocks should use the revision ID and cache forever. 

CleanShot 2023-05-28 at 19 05 19@2x


routes/_middleware.ts
```diff
@@ -86,7 +88,19 @@ export const handler = async (
...
+    url.searchParams.has("asRaw") &&
+    newHeaders.get("content-type")?.includes("text/html")
+  ) {
+    const html = await initialResponse.text();
+    const parser = new DOMParser();
+    const doc = parser.parseFromString(html, "text/html");
+    const firstChild = doc?.querySelector("body")?.firstElementChild;
+    firstChild?.querySelector("script")?.remove();
+    bodyReadable = firstChild?.outerHTML ?? null;
+  }
```
mcandeia, line 101: nit: You can do it later.

I'd rather prefer put this on [fresh handler](https://github.com/deco-cx/live.ts/blob/main/handlers/fresh.ts#L24) (together along `asJson`) just for simplicity and localization.

PS: There's no overhead in copying headers twice

mcandeia (approved):
@igorbrasileiro can you take a look as well?

PR #296, opened 2023-06-13
feat: Handle formdata at preview

igorbrasileiro:
also drop layout props

### DIFF
--- a/routes/live/previews/[...block].tsx
+++ b/routes/live/previews/[...block].tsx
@@ -50,6 +50,26 @@ export default function Preview(props: PageProps<Page>) {
 
 const addLocal = (block: string): string =>
   block.startsWith("islands") && block.endsWith("tsx") ? `./${block}` : block;
+
+const getPropsFromRequest = async (req: Request) => {
+  const url = new URL(req.url);
+  if (req.method === "POST") {
+    const reader = req.body?.getReader();
+    reader?.releaseLock();
+    const data = (req.headers.get("content-type") ===
+        "application/x-www-form-urlencoded"
+      ? JSON.parse(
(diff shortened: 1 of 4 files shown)


routes/live/previews/[...block].tsx
```diff
@@ -50,6 +50,26 @@ export default function Preview(props: PageProps<Page>) {
 
 const addLocal = (block: string): string =>
   block.startsWith("islands") && block.endsWith("tsx") ? `./${block}` : block;
+
+const getPropsFromRequest = async (req: Request) => {
+  const url = new URL(req.url);
+  if (req.method === "POST") {
+    const reader = req.body?.getReader();
+    reader?.releaseLock();
```
Comment (mcandeia) [inline routes/live/previews/[...block].tsx:58]: this is necessary to be done twice? I think you can let only the last one

routes/live/previews/[...block].tsx
```diff
@@ -50,6 +50,26 @@ export default function Preview(props: PageProps<Page>) {
...
+    const data = (req.headers.get("content-type") ===
```
Comment (mcandeia) [inline routes/live/previews/[...block].tsx:59]: extract content-type to a const

mcandeia (approved):

routes/live/previews/[...block].tsx
Comment (igorbrasileiro) [reply routes/live/previews/[...block].tsx:58]: No, can't because when call getReader, the reader is locked: https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/getReader

Comment (igorbrasileiro) [reply routes/live/previews/[...block].tsx:58]: image

Comment (igorbrasileiro) [reply routes/live/previews/[...block].tsx]: I dropped both

PR #303, opened 2023-06-19
fix CORS error of /previews when searchParams has mode=showcase

marialuizaaires:


routes/_middleware.ts
```diff
@@ -73,8 +73,13 @@ export const handler = async (
 
   // Let rendering occur — handlers are responsible for calling ctx.state.loadPage
   const initialResponse = await ctx.next();
-
   const newHeaders = new Headers(initialResponse.headers);
+  if (
+    url.pathname.startsWith("/live/previews") &&
+    url.searchParams.has("mode") && url.searchParams.get("mode") == "showcase"
+  ) {
+    Object.entries(allowCorsFor(req)).map(([x, y]) => {newHeaders.set(x, y)})
```
matheusgr, line 81:
```suggestion
    Object.entries(allowCorsFor(req)).map(([name, value]) => {newHeaders.set(name, value)})
```

matheusgr (approved):

PR #313, opened 2023-06-22
fix: make dot work again

tlgimenes:
Adds PathTemplate to the querystring so when pressing dot everythign works again


mcandeia (approved):
Salve!!

PR #314, opened 2023-06-23
Long live to hints! No more max-depth on resolve & 3x faster

mcandeia:
image

A lot of improvements on the resolution algorithm (added on README), also we've got performance improvements!

Aside from that, this PR includes:

1. A new `onBeforeResolveProps` function that any block can export to "intercept" the resolution algorithm. This was used for audiences to maintain backwards compatibility with the current version where the `resolve depth` prevents to resolve the page path.
2. Options to decide wether the result can be null if dangling
3. Verify the routes `destination` to skip the resolution algorithm to static and internal routes (previously it was necessary because it was ambiguous if it is a proxy or the current static route)
4. New tests and bench tests
5. Removed support to resolve objects when returned by resolvers
6. Fixed invoke props which was leveraging the bug-feature of resolve-depth being maximum at 3


import_map.json
```diff
@@ -10,7 +10,7 @@
     "https://deno.land/std@0.150.0/": "https://deno.land/std@0.181.0/",
     "https://deno.land/std@0.152.0/": "https://deno.land/std@0.181.0/",
     "https://deno.land/std@0.164.0/": "https://deno.land/std@0.181.0/",
-    "$fresh/": "https://deno.land/x/fresh@1.1.4/",
+    "$fresh/": "https://denopkg.com/deco-cx/fresh@1.3.0/",
```
lucis, line 13: out ouf curiosity, why did you need that?

lucis (approved):

import_map.json
mcandeia: It was outdated with two things

1. Now middlewares receives the `destination` property which tells if the next route will be either internal, static or a "normal" route, then I could use it to skip config resolution
2. the `config` property was not being used from fresh on manifest (which was causing some checks to fail)

PR #321, opened 2023-06-27
Allow users to create variants of their sections in a page

mcandeia:


matheusgr (approved):

blocks/matcher.ts
```diff
@@ -50,7 +79,7 @@ export interface MatcherModule extends
     boolean | ((ctx: MatchContext) => boolean),
     (ctx: MatchContext) => boolean
   > {
-  unstable?: boolean;
+  stickness?: boolean;
```
mcandeia, line 82: Stickiness*

PR #345, opened 2023-07-22
Add a newline when writing stringified JSONs

victorsenam:
While developing a site I've realised that `deno fmt` adds a newline to end of every tracked git file. Since the logic here does not add newlines to end of JSON.strinfigy before adding it to a file, it is not following the format commited to the repository.

This is a small nuisance and since I'm trying things out decided to try and fix it.

Before:
- Clone a repo which uses deco
- Run `deno task start`
- Run `git diff` or `git status` and notice that there's a whitespace removed at the end of `schemas.gen.json`
- Run `deno fmt`
- Run `git diff` or `git status` and notice that such space is gone

After:
- Clone a repo which uses deco
- Run `deno task start`
- Run `git diff` or `git status` and see that there's no change

The video demonstrates those versions:

https://github.com/deco-cx/deco/assets/3386440/f5872794-b379-4632-9fcd-7e5af4d680f3

### DIFF
--- a/utils/stringifyForWrite.ts
+++ b/utils/stringifyForWrite.ts
@@ -1,0 +1,5 @@
+export default function stringifyForWrite(
+  object: Object,
+): string {
+  return `${JSON.stringify(object, null, 2)}\n`;
+}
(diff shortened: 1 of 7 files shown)


utils/stringifyForWrite.ts
```diff
@@ -0,0 +1,5 @@
+export default function stringifyForWrite(
```
mcandeia, line 1: would you mind to rename it to `utils/json.ts` ?

mcandeia (approved):
Thanks!!

LGTM

PR #355, opened 2023-07-28
App block

mcandeia:


engine/schema/utils.ts
```diff
@@ -167,11 +161,20 @@ function isQuotaExceededError(err: unknown): boolean {
...
   path: string,
-  importMap?: string,
+  _importMap?: string,
 ): Promise<DocNode[]> => {
   try {
+    const docCacheKey = path;
+    if (context.isDeploy) {
+      const cached = denoDocLocalCache[docCacheKey];
+      if (!cached) {
+        throw new Error(`could not resolve ${docCacheKey} on denodoc`);
```
matheusgr, line 174: Offer a solution for the probably error cause.

```suggestion
        throw new Error(`could not resolve ${docCacheKey} on denodoc`); Create denodoc cache locally with "deno task doccache" and push the denocache.zst file.
```

matheusgr: Do we have a task to create the deno doc cache?

blocks/app.ts
```diff
@@ -0,0 +1,143 @@
...
+  TAppManifest extends AppManifest = AppManifest,
+  // deno-lint-ignore ban-types
+  TAppState = {},
+> {
+  manifest: TAppManifest;
+  state: TAppState;
+}
+
+export type AppContext<TApp extends App = App> = FnContext<
+  TApp["state"],
```
matheusgr, line 32: Maybe `TApp["state"]` should merge its own state and those from its dependencies?

Something like that may be better modeled as Record<namespace, {}>.

matheusgr (changes requested):

engine/schema/utils.ts
mcandeia: yes. deno task start --gen-only

blocks/app.ts
mcandeia: Discussed offline

scripts/apps/init.ts
```diff
@@ -0,0 +1,92 @@
...
+
+const templates: Templates = ["import_map.json", "deno.json", "deco.ts", "state.ts", "deps.ts", {
+  app: {
+    loaders: ["bin.ts"],
+  },
+}];
+
+const init = async () => {
+  const latestVersionPromise = lookup(
+    "https://denopkg.com/deco-cx/deco@main/",
```
matheusgr, line 73: Don't see main as a common version name/tag.

```suggestion
    "https://denopkg.com/deco-cx/deco@latest/",
```

matheusgr (approved):

PR #356, opened 2023-07-29
[fix] Rename utils/json.ts and improve typing

victorsenam:


utils/json.ts
```diff
@@ -1,5 +1,5 @@
-export default function stringifyForWrite(
-  object: Object,
+export function stringifyForWrite(
+  object: object,
```
victorsenam, line 2: linter told me to use this `object` instead of the `Object`. :)

mcandeia (approved):

PR #372, opened 2023-08-10
Add websocket renderer

pedroteosousa:
This will be used by admin on the Visual Library


routes/live/previews/[...block].tsx
```diff
@@ -55,10 +55,40 @@ export const handler = async (
...
+    cache[key] ??= render(data.url, props, req, ctx).then((response) => response.text().then(function (html) {
+      const response = JSON.stringify({
+        html,
+        key,
+      });
+      return response;
+    }));
+    socket.send(await cache[key]);
+  };
+  socket.onclose = () => console.log("render socket closed.");
```
Comment (mcandeia) [inline routes/live/previews/[...block].tsx:80]: Use ctx.state.log instead so we can activate when necessary

mcandeia (approved):
Replace all console.logs to ctx.state.log and ship it

PR #375, opened 2023-08-12
remove unsupported sintaxe at safari 12 or less

guitavano:
Apparently, safari 12 or less does not support syntaxes as "??" or "?."


guitavano:
Resolved

PR #438, opened 2023-09-17
Fix missing sw.js on upgrade and missing 'deco/' alias on update

IncognitaDev:


mcandeia (approved):

utils/update.ts
```diff
@@ -53,28 +51,28 @@ export async function update() {
...
         importMap.imports[pkg] = url.at(latestVersion).url;
       }
 
       return importMap;
     },
-    Promise.resolve(Object.assign({ imports: {} }, importMap)),
+    Promise.resolve(Object.assign({ imports: {} }, importMap))
   );
 
+  if (!newImportMap.imports["deco/"] && newImportMap.imports["$live/"]) {
```
mcandeia, line 65: Check this out => https://github.com/deco-cx/deco/blob/main/scripts/upgrade.ts#L408

mcandeia (changes requested):

utils/update.ts
IncognitaDev: Its different scripts and in both "deco/' alias is required. Anyway, in update script it will be add just if its doesnt exists and in upgrade script its be add using a JS obejct, wich means any duplicated key will be override.

mcandeia (approved):

PR #451, opened 2023-09-27
Fix allocate blocks and otel parentbased tracer

mcandeia:


observability/otel/config.ts
```diff
@@ -56,13 +56,15 @@ const tracingSampleRatio = Deno.env.has(OTEL_TRACING_RATIO_ENV_VAR)
...
 const provider = new NodeTracerProvider({
   resource: resource,
   sampler: new ParentBasedSampler(
     {
-      root: new TraceIdRatioBasedSampler(tracingSampleRatio),
-      localParentSampled: debugSampler,
-      localParentNotSampled: debugSampler,
+      root: debugSampler,
+      localParentNotSampled: traceIdRatioBasedSampler,
     },
```
matheusgr, line 68: Maybe we should use a custom sampler... (extend and override shouldSample from https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-sdk-trace-base/src/sampler/TraceIdRatioBasedSampler.ts)

rationale: setting debugSampler may create a record decision at top level, even if the span is not sampled by the traceIdRatioBasedSampler.

matheusgr (approved):

PR #460, opened 2023-10-02
feat: add response status editable by loader

IncognitaDev:
When a search return 0 products, some SEO agencies demand the page return to be 404 as response status.
In antive vtex stores not found PDPs return 404 as a reponse status (legacy and storefront)

This PR add Status in ctx to be editable by loader.

Use Case: 

Do a invalid search
in inline loader, check the product return length
if 0, change the status to 404 with `ctx.response.status`

### DIFF
--- a/routes/_middleware.ts
+++ b/routes/_middleware.ts
@@ -18,7 +18,8 @@ import { formatLog } from "../utils/log.ts";
 export const redirectToPreviewPage = async (url: URL, pageId: string) => {
   url.searchParams.append("path", url.pathname);
   url.searchParams.set("forceFresh", "");
-  if (!url.searchParams.has("pathTemplate")) { // FIXM(mcandeia) compatibility mode only, once migrated pathTemplate is required because there are pages unpublished
+  if (!url.searchParams.has("pathTemplate")) {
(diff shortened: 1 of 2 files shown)


routes/_middleware.ts
```diff
@@ -48,171 +49,179 @@ const isMonitoringRobots = (req: Request) => {
...
+        newHeaders.set(name, value);
+      });
+    }
+    response.headers.forEach((value, key) => newHeaders.append(key, value));
+    const printTimings = ctx?.state?.t?.printTimings;
+    printTimings && newHeaders.set("Server-Timing", printTimings());
+
+    if (
+      url.pathname.startsWith("/_frsh/") &&
+      [400, 404, 500].includes(initialResponse.status)
```
mcandeia, line 200: I suggest you to extract the variable `status` and reuse it here to keep consistency across usages.

mcandeia (changes requested):
Can you provide a site link which I can assure that the behavior is working properly?

IncognitaDev:
> Can you provide a site link which I can assure that the behavior is working properly?

I dont know how to link a fork in my site but i have this repo with a example implemented, you can clone and link this deco version to view the behaviour .

https://github.com/deco-sites/vtexlegacydebug

mcandeia (approved):

PR #472, opened 2023-10-11
Allow admin.deco.cx to preview

lucis:
image

### DIFF
--- a/utils/admin.ts
+++ b/utils/admin.ts
@@ -1,6 +1,7 @@
 import type { JSONSchema7 } from "../deps.ts";
 
 export const adminDomain = `https://deco.cx`;
+export const adminAlternateDomain = `https://admin.deco.cx`;
 export const landingPageDomain = `https://www.deco.cx`;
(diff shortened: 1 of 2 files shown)


utils/admin.ts
```diff
@@ -1,6 +1,7 @@
 import type { JSONSchema7 } from "../deps.ts";
 
 export const adminDomain = `https://deco.cx`;
+export const adminAlternateDomain = `https://admin.deco.cx`;
```
igorbrasileiro, line 4:
```suggestion
const adminDomains = [adminDomain, "https://admin.deco.cx"];
```

igorbrasileiro: edit line 10 to be 
```ts
  if (adminDomains.some((_adminDomain) => url.startsWith(_adminDomain))) {

```

igorbrasileiro (approved):

utils/admin.ts
lucis: We use `adminDomain` to create links as well. I think it's ok

PR #475, opened 2023-10-17
Add supabase logs

mcandeia:


engine/releases/supabaseProvider.ts
```diff
@@ -134,10 +138,11 @@ export const newSupabase = (
...
       }, (_status, err) => {
         if (err) {
-          console.error(
-            "error when trying to subscribe to release changes falling back to background updates",
-            err,
-          );
+          const errMsg =
+            `error when trying to subscribe to release changes falling back to background updates, ${err}`;
+          logger.error(errMsg);
+          console.error(errMsg);
```
lucis, line 144: Begs the question if we need this

lucis (approved):

engine/releases/supabaseProvider.ts
mcandeia: yes, because somehow the logger.error interface is different from the console.error

PR #479, opened 2023-10-24
feat: nested partials

tlgimenes:
This PR moves the partials feature into deco. Read the upgraded docs to understand how to use it:
https://github.com/deco-sites/starting/pull/193


engine/core/resolver.ts
```diff
@@ -87,6 +87,41 @@ export type ResolvesTo<
...
+
+  for (let it = 0; it < minified.length; it += 2) {
+    unwinded.push({
+      type: RESOLVE_TYPES[Number(minified[it])],
+      value: minified[it + 1],
+    });
+  }
+
+  return unwinded;
+};
```
mcandeia, line 123: nit: I prefer the pattern that you can create a "build" object which encapsulates both logic

const FieldResolver = {
minify: (chain: FieldResolver[]) => {},
unwind: (minified: string[]) => {}
}

engine/manifest/manifest.ts
```diff
@@ -183,7 +189,7 @@ export const createResolver = <
 
   context.release = provider;
   const resolver = new ReleaseResolver<FreshContext>({
-    resolvers: { ...resolvers, ...defaultResolvers },
+    resolvers: { ...resolvers, ...defaultResolvers, ...freshResolvers },
```
mcandeia, line 192: can you check if anywhere that is referencing defaultResolvers is also referencing freshResolvers now?

try to find for all "defaultResolvers" occurrence throughout the project and check if the fresh resolvers should be included there as well

mcandeia (approved):

PR #482, opened 2023-10-24
feat: support deno.jsonc

lui-dias:
Add support to `deno.jsonc` on `deno run -Ar https://deco.cx/update`


scripts/update.ts
```diff
@@ -1,90 +1,114 @@
-import * as semver from "https://deno.land/x/semver@v1.4.1/mod.ts";
-import {
-  lookup,
-  REGISTRIES,
-} from "https://denopkg.com/hayd/deno-udd@0.8.2/registry.ts";
-import { parse } from "https://deno.land/std@0.204.0/flags/mod.ts";
-import { join } from "https://deno.land/std@0.204.0/path/mod.ts";
-import { stringifyForWrite } from "../utils/json.ts";
+import * as semver from 'https://deno.land/x/semver@v1.4.1/mod.ts'
```
mcandeia, line 1: are you using tabs for for indentation? can you fix to use 2 spaces instead?

scripts/update.ts
```diff
@@ -1,90 +1,114 @@
...
-    await Deno.readTextFile(importMapPath).then(JSON.parse),
-    importMapPath,
-  ];
-};
+const getImportMap = async (dir: string): Promise<[ImportMapFile, string]> => {
+	let denoJsonFile = ''
+
+	if (await exists('deno.json')) {
+		denoJsonFile = 'deno.json'
+	} else if (await exists('deno.jsonc')) {
```
mcandeia, line 32: you're using single quotes, can you change to use double quotes instead?

Also: I think you should `join(dir, "deno.json")` to use `exists` right?

mcandeia (changes requested):

scripts/update.ts
lui-dias: Yep, I forgot to join. I made the commit with the fix

PR #491, opened 2023-10-30
Allow fetching release only having the URL

mcandeia:
The main idea of this PR is letting DECO_RELEASE environment var being opaque. Now, you can set any url that can be fetched through regular http fetch and the data will be used as the current release. This made cloudflare provider obsolete as this can be converted to a http fetch to https://configs.decocdn.com/forrageira/a30ca6dd-0170-418e-8017-f5a791b4c85f.json for instance


engine/manifest/defaults.ts
```diff
@@ -190,7 +190,7 @@ export default {
       if (!(err instanceof HttpError)) {
         throw new HttpError(
           new Response(
-            err ? JSON.stringify(err) : JSON.stringify({
+            err ? err : JSON.stringify({
```
mcandeia, line 193: minor fix please let it pass rsrs

tlgimenes (approved):

PR #502, opened 2023-11-21
feat: Caching loaders

tlgimenes:
# Purpose of this PR

This pull request introduces enhancements to loaders by incorporating stale caching and improved visibility. 

# Motivation

The architecture of Deco.cx relies on the assumption that utilizing stale caching for third-party datasources yields optimal performance and facilitates customization. In practical terms, our custom `fetch` function ensures that all requests are routed through a stale caching CDN layer. For example, when a user writes `fetch('https://example.com')`, Deco.cx's framework actually fetches the proxied URL: `fetch('https://decocache.com?src=https://example.com')`. While this approach initially worked well for integrations on `decohub`, challenges arose as some services offered their own client packages that did not support a custom fetch. Additionally, instrumenting these fetches for effective tracing in our admin UI proved to be a hurdle.

To address these issues, we are shifting away from a custom `fetch` approach and pivoting towards enhancing our loader API to support caching and tracing.

# Opting into the New Features

To adopt the updated API, set the following environment variable:

```sh
source ENABLE_LOADER_CACHE=true
```

And that's it! You are now leveraging the new traced loaders.

## Stale Caching Mode

By default, stale caching is inactive in loaders. To activate stale caching mode, modify your loader's code as follows:

```ts
// ... loader code

export const cache = 'stale-while-revalidate'
```

The `cache` variable can be set to either `stale-while-revalidate` or `no-store`, with the default being `no-store`.

Now, your loader is stale cached! To confirm, open a page where your loader is used and add a `?__d` query string to the URL. You should observe the tracing information in the terminal, including cache states such as `HIT`, `MISS`, `STALE`, and `BYPASS`.

- `HIT`: Cache considered fresh. We use 60 seconds as the standard max-age for caching
- `STALE`: Cache not fresh, but stale content served (background validation triggered).
- `MISS`: Loader is cacheable, but data not in the cache.
- `BYPASS`: Loader not cacheable, possibly due to using `export const cache = 'no-store'` mode, missing environment variable, or no caching layer available.

## Varying the Cache

Some loaders may need to vary based on properties available only in the request object, such as query strings. To achieve this, use the new `cacheKey` function:

```ts
export const cacheKey = (req: Request, ctx: AppContext): string => '';
```

Simply export this function in your loader file. An example can be found [here](https://github.com/deco-cx/apps/blob/333c1d4c74493b18762cc3c1abd1f0d929431dc5/vtex/loaders/intelligentSearch/productDetailsPage.ts#L130).

### DIFF
--- a/blocks/loader.ts
+++ b/blocks/loader.ts
@@ -14,6 +14,7 @@ import {
   FnProps,
   SingleFlightKeyFunc,
 } from "./utils.tsx";
+import { logger } from "deco/observability/otel/config.ts";
 
 export type Loader = InstanceOf<typeof loaderBlock, "#/root/loaders">;
 
@@ -84,13 +85,15 @@ const stats = {
   }),
 };
 
-let maybeCache: Promise<unknown> | Cache | undefined = caches.open("loader")
+let maybeCache: Cache | undefined;
+
+caches.open("loader")
   .then((c) => maybeCache = c)
   .catch(() => maybeCache = undefined);
 
 const MAX_AGE_S = 60; // 60 seconds
 
-const isCache = (c: any): c is Cache => typeof c?.put === "function";
+const isCache = (c: Cache | undefined): c is Cache => typeof c !== "undefined";
 
 const inFuture = (maybeDate: string) => {
   try {
@@ -102,6 +105,12 @@ const inFuture = (maybeDate: string) => {
 
 const noop = () => "";
 
+/**
+ * Wraps the loader written by the user by adding support for:
+ * 1. Caching
+ * 2. Single Flight
+ * 3. Tracing
+ */
 const wrapLoader = ({
   default: handler,
   cache: mode = "no-store",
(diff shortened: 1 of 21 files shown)


blocks/loader.ts
```diff
@@ -57,24 +68,145 @@ export const wrapCaughtErrors = async <
...
+          stats.cache.add(1, { status, loader });
+
+          return await handler(props, req, ctx);
+        }
+
+        // Somehow typescript does not understand maybeCache is Cache
+        const cache = maybeCache as Cache;
+
+        // TODO: Resolve props cache key statically
+        const key = `${hash(props)}-${cacheKey(req, ctx)}`;
```
mcandeia, line 146: can you add a metric to meter how long this is taking?

blocks/loader.ts
```diff
@@ -57,24 +68,145 @@ export const wrapCaughtErrors = async <
...
+        const request = new Request(
+          `https://localhost?propsKey=${hash(props)}&requestKey=${key}`,
```
mcandeia, line 148: add a comment explaning why localhost (I know that its just an arbitrary cache key and the host doesn't matter but it might worth to let it explicit)

blocks/loader.ts
```diff
@@ -57,24 +68,145 @@ export const wrapCaughtErrors = async <
...
+
+          cache.put(
+            request,
+            new Response(JSON.stringify(json), {
+              headers: {
+                "expires": new Date(Date.now() + (MAX_AGE_S * 1e3))
+                  .toUTCString(),
+              },
+            }),
+          ).catch(console.error);
```
mcandeia, line 162: should we meter it? I think you can use `logger` instead of `console.error`

blocks/loader.ts
```diff
@@ -57,24 +68,145 @@ export const wrapCaughtErrors = async <
...
+          }
+
+          const expires = matched.headers.get("expires");
+          const isStale = expires ? !inFuture(expires) : false;
+
+          if (isStale) {
+            status = "stale";
+            stats.cache.add(1, { status, loader });
+
+            callHandlerAndCache().catch((error) => console.error(error));
```
mcandeia, line 184: logger instead of console.error

blocks/loader.ts
```diff
@@ -57,24 +68,145 @@ export const wrapCaughtErrors = async <
...
+  try {
+    return new Date(maybeDate) > new Date();
+  } catch {
+    return false;
+  }
+};
+
+const noop = () => "";
+
+const wrapLoader = ({
```
mcandeia, line 105: it might worth to have a comment for this function

mcandeia (changes requested):

blocks/loader.ts
```diff
@@ -57,24 +68,145 @@ export const wrapCaughtErrors = async <
...
+
+      try {
+        if (skipCache) {
+          status = "bypass";
+          stats.cache.add(1, { status, loader });
+
+          return await handler(props, req, ctx);
+        }
+
+        // Somehow typescript does not understand maybeCache is Cache
```
igorbrasileiro, line 142: With the change i proposed above, the cache is `undefined | cache`.

runtime/caches/denoKV.ts
```diff
@@ -261,48 +205,54 @@ export const caches: CacheStorage = {
...
+          for (const chunk of await chunks) {
+            length += chunk.value?.length ?? 0;
+          }
+        }
+
+        const result = new Uint8Array(length);
+
+        let bytes = 0;
+        for (const chunks of many) {
+          for (const chunk of await chunks) {
```
igorbrasileiro, line 244:
```suggestion
const chunks = await Promise.all(many.flat(Infinity));
for (const chunk of chunks) {
```

igorbrasileiro (approved):

mcandeia (approved):

PR #531, opened 2024-01-04
Loader cache start treshold

matheusgr:


blocks/loader.ts
```diff
@@ -74,6 +75,12 @@ export const wrapCaughtErrors = async <
 export const ENABLE_LOADER_CACHE =
   Deno.env.get("ENABLE_LOADER_CACHE") !== undefined;
 
+export const LOADER_CACHE_START_TRESHOLD =
+  Deno.env.get("LOADER_CACHE_START_TRESHOLD") ?? 5;
+
+export const LOADER_CACHE_SIZE =
+  Deno.env.get("LOADER_CACHE_SIZE") ?? 1_024;
+  1_024
```
hugo-ccabral, line 83: is this correct?

hugo-ccabral (commented):
nit

blocks/loader.ts
matheusgr: its a random wild number misplaced :)

guitavano (approved):

PR #545, opened 2024-01-16
instruments the put at proxy's cache

hugo-ccabral:


runtime/caches/proxy.ts
```diff
@@ -116,7 +117,35 @@ export const caches: CacheStorage = {
...
+          const length = response.headers.get("content-length");
+          if (!length) {
+            const responseClone = response.clone();
+            return (await responseClone?.text()).length ?? 0;
+          }
+          return Number(length);
+        };
+
+        const url = req.url;
+        const responseLength = await getResponseLength(response);
```
matheusgr, line 132: can you only run this code if you are executing a trace?

PR #546, opened 2024-01-17
Ft/new hash algo

hugo-ccabral:
Hashes with a new algorithm

Benchmark with a random object (generated by copilot):
image


blocks/loader.ts
```diff
@@ -1,7 +1,7 @@
 // deno-lint-ignore-file no-explicit-any
-import hash from "https://esm.sh/v135/object-hash@3.0.0";
 import JsonViewer from "../components/JsonViewer.tsx";
 import { ValueType } from "../deps.ts";
+import crypto from "node:crypto";
```
mcandeia, line 4: you can use crypto without importing from node

mcandeia: actually deno has it's own crypto implementation

https://deno.land/api@v1.39.4?s=Crypto

benchs/hash.ts
```diff
@@ -0,0 +1,80 @@
+import { webcrypto } from "node:crypto";
```
mcandeia, line 1: you can create .bench.ts files alongside the file that exports your function that you're going to test. I think this is the standard adopted by deno community

mcandeia (changes requested):

blocks/loader.ts
hugo-ccabral: but this module is just to create randoms

hugo-ccabral: it is not stable, as my current implementation, so I'm just optimizing the previous approach

blocks/loader.bench.ts
```diff
@@ -0,0 +1,93 @@
...
+//     for (const test of tests) {
+//         await md5(JSON.stringify(test));
+//     }
+// });
+
+// Deno.bench("hash-wasm sha1", { group: "object-hash" }, async () => {
+//     for (const test of tests) {
+//         await sha1(JSON.stringify(test));
+//     }
+// });
```
matheusgr, line 93:
```suggestion
```

matheusgr (approved):

blocks/loader.ts
```diff
@@ -192,11 +192,16 @@ const wrapLoader = ({
...
+              ignoreUnknown: true,
+              respectType: false,
+              respectFunctionProperties: false,
+              algorithm: "md5",
+            });
+            span.setAttributes({
+              hash_size_bytes: hashedProps.length * 2,
+              props_length: JSON.stringify(props).length,
+            });
             url.searchParams.set("props", hashedProps);
```
matheusgr, line 205: could this not run if otel is not active? if so, this is a bug. it should always run set("props").

matheusgr (approved):

engine/schema/schemeable.test.ts
```diff
@@ -77,7 +77,7 @@ Deno.test("Simple type generation", async () => {
     fail("SimpleType should exists");
   }
 
-  const name = Deno.build.os === "windows" ? "tl@157-178" : "tl@155-174";
+  const name = "tl@155-174";
```
mcandeia, line 80: Will this pass even in different OSs?

mcandeia (approved):

engine/schema/schemeable.test.ts
hugo-ccabral: I'm fixing now so don't break the build. But my check is broken locally, even with my Deno.build.os returning "windows"

blocks/loader.ts
hugo-ccabral: no, it will always run. Even if otel disabled

PR #551, opened 2024-01-25
remove treshold

tlgimenes:


guifromrio (approved):

guifromrio:
@tlgimenes @matheusgr @mcandeia couldn't this be configurable via admin? so we can play around with the values dynamically

PR #555, opened 2024-01-31
replace object-hash when resolveChain and revision id are available

hugo-ccabral:


blocks/loader.ts
```diff
@@ -190,31 +200,18 @@ const wrapLoader = ({
...
-          });
-        } catch (e) {
-          span.recordException(e);
-          throw e;
-        } finally {
-          timing?.end();
-          span.end();
+        const resolveChainString = FieldResolver.minify(resolveChain)
+          .toString();
+        const revisionID = await release?.revision() ?? undefined;
```
mcandeia, line 205: import { Context } from "deco/deco.ts";
...
// within the func
const context = Context.active();
const release = context.release;
//

the release changes based on the context

mcandeia (changes requested):

mcandeia (approved):

PR #576, opened 2024-02-19
Try a stable manifest gen approach

mcandeia:


viktormarinho:
Working on both Macos & pop_OS

engine/manifest/manifestGen.ts
```diff
@@ -84,6 +84,8 @@ export const decoManifestBuilder = async (
     let totalBlocks = 0;
     const blockDir = join(dir, blk.type);
 
+    // for stabilizy purposes we need to sort paths first.
```
viktormarinho, line 87:
```suggestion
    // for stability purposes we need to sort paths first.
```

PR #579, opened 2024-02-22
Force https on request passed by engine

mcandeia:


runtime/caches/fileSystem.ts
```diff
@@ -0,0 +1,216 @@
...
+        },
+      });
+    },
+  };
+
+  // Check if the cache directory exists, if not, create it
+  if (FILE_SYSTEM_CACHE_DIRECTORY && !existsSync(FILE_SYSTEM_CACHE_DIRECTORY)) {
+    Deno.mkdirSync(FILE_SYSTEM_CACHE_DIRECTORY, { recursive: true });
+  }
+  return caches;
```
mcandeia, line 211: we can't do that at module-level. You need to do it lazy, also, we should have a way to check if we have write permissions and then fallback to undefined

ItamarRocha: Done

PR #591, opened 2024-03-01
fix: shortcircuit doesnt works

lui-dias:
throw error from `shortcircuit` could not be detect because of async rendering


blocks/section.ts
```diff
@@ -17,6 +17,7 @@ import {
 import { Resolver } from "../engine/core/resolver.ts";
 import { AppManifest, FunctionContext } from "../types.ts";
 import { PartialProps } from "$fresh/src/runtime/Partial.tsx";
+import { HttpError } from "deco/engine/errors.ts";
```
mcandeia, line 20: Use relative import instead

mcandeia (approved):
Fix then shipit

PR #595, opened 2024-03-05
create hook useDevice()

guitavano:


mcandeia (changes requested):
I understand the rational behind this change and I know that is sucks to create an inline loader just to access device 90% of the times that you need a loader. However I think this is beneficial for many reasons, I will try to list some of them here but this is not exhaustive, just few of them that I'm thinking right now:

1) It is less react-compatible component

Meaning that reusability becomes difficult, you can't export this section as an Island, for instance, because it now receives request and for the most of cases request is not available to be used, same for device.

2) Component props now can receive request and device, which would need us to create extra features to hide/ignore them when generating schema.

3) users can do it by themself in userland using an app middleware, we can improve the ux for doing it to be a brainless operation

4) A better alternative more react-compatible can be using hooks? I know there are scenarios where hooks won't work like inside islands that runs client-side, but well our users is facing these problems today

guitavano:
I followed Candeia instructions and created a Hook, and better news! Hooks with section context now works on islands, so let's gooo.

PR #596, opened 2024-03-05
feat: schema generation handle chose function return types as array

igorbrasileiro:
This fixes this problem:

scenario
```ts
// MyCustomSection.tsx
type MyCustomSection = JSX.Element;
interface Props {}
export function loader(p: Props) {}
export default function Section(): MyCustomSection {}
```

```ts
// some block that only receives section<MyCustomSection>;
import { Section } from "deco/blocks/section.ts";

interface Props {
  section: Section<MyCustomSection>
}
export default function Block(p: Props) {}
```

Before:
the schema of property section of Block will be an array with { title: "MyCustomSection" }

After:
the schema of property section of block will be an array with {title: "MyCustomSection"} and schema of props from `MyCustomSection.tsx`

### DIFF
--- a/blocks/section.ts
+++ b/blocks/section.ts
@@ -105,7 +105,7 @@ export const createSectionBlock = (
   type: "sections" | "pages",
 ): Block<SectionModule> => ({
   type,
-  introspect: { funcNames: ["loader", "default"], includeReturn: true },
+  introspect: { funcNames: ["loader", "default"], includeReturn: ["default"] },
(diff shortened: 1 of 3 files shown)


mcandeia (approved):

lucis (approved):
Nice!

Are there any test files in this part of the code?

mcandeia:
> Nice!
> 
> Are there any test files in this part of the code?

Not about sections specifically, only schema generation overall

PR #600, opened 2024-03-08
Tiered cache

ItamarRocha:
This PR created the tiered cache. A cache that can depend on multiple caches and offer a fallback mechanism between them.

Improvements with this PR:
* Avoid cache cold start when creating a new deploy.
* Make executing the requests the last resort; prefer going and checking in each cache available (way faster).
* Faster websites.

Example:
https://sites-osklenbr-onozcgfeuy.decocdn.com/masculino-casacos?__d=trace-tiered

### DIFF
--- a/runtime/caches/fileSystem.ts
+++ b/runtime/caches/fileSystem.ts
@@ -62,6 +62,8 @@ function createFileSystemCache(): CacheStorage {
       const fileContent = await Deno.readFile(filePath);
       return fileContent;
     } catch (err) {
+      // Error code different for file/dir not found
+      // The file won't be found in cases where it's not cached
       if (err.code === "ENOENT") {
         logger.warning(
           `file not found when reading from file system, path: ${FILE_SYSTEM_CACHE_DIRECTORY}/${key}`,
@@ -105,7 +107,6 @@ function createFileSystemCache(): CacheStorage {
     },
     open: (cacheName: string): Promise<Cache> => {
       const requestURLSHA1 = withCacheNamespace(cacheName);
-
       return Promise.resolve({
         /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Cache/add) */
         add: (_request: RequestInfo | URL): Promise<void> => {
@@ -154,10 +155,8 @@ function createFileSystemCache(): CacheStorage {
             span.addEvent("file-system-get-data");
 
             if (data === null) {
-              span.addEvent("cache-miss");
               return undefined;
             }
-            span.addEvent("cache-hit");
 
             downloadDuration.record(downloadDurationTime, {
               bufferSize: data.length,
@@ -167,7 +166,6 @@ function createFileSystemCache(): CacheStorage {
               data,
             );
           } catch (err) {
-            span.recordException(err);
             throw err;
           } finally {
             span.end();
@@ -242,6 +240,7 @@ const hasWritePerm = async (): Promise<boolean> => {
   ).then((status) => status.state === "granted");
 };
 
-export const caches = await hasWritePerm() && FILE_SYSTEM_CACHE_DIRECTORY
-  ? createFileSystemCache()
-  : undefined;
+export const isFileSystemAvailable = await hasWritePerm() &&
+  FILE_SYSTEM_CACHE_DIRECTORY !== undefined;
+
+export const caches = createFileSystemCache();
--- a/runtime/caches/mod.ts
+++ b/runtime/caches/mod.ts
@@ -1,10 +1,16 @@
 import { withInstrumentation } from "./common.ts";
-import { compose } from "./compose.ts";
 import { caches as cachesKV } from "./denoKV.ts";
-import { caches as cachesProxy } from "./proxy.ts";
+import {
+  caches as cachesFileSystem,
+  isFileSystemAvailable,
+} from "./fileSystem.ts";
 import { caches as redisCache, redis } from "./redis.ts";
+import { caches as cachesS3, isS3Available } from "./s3.ts";
+import { createTieredCache } from "./tiered.ts";
 
-const DEFAULT_CACHE_ENGINE: CacheEngine = "CF_PROXY";
+export const ENABLE_LOADER_CACHE =
+  Deno.env.get("ENABLE_LOADER_CACHE") === "true";
+const DEFAULT_CACHE_ENGINE = "KV";
 const WEB_CACHE_ENGINES: CacheEngine[] = Deno.env.has("WEB_CACHE_ENGINE")
   ? Deno.env.get("WEB_CACHE_ENGINE")!.split(",") as CacheEngine[]
   : [DEFAULT_CACHE_ENGINE];
@@ -14,7 +20,13 @@ export interface CacheStorageOption {
   isAvailable: boolean;
 }
 
-export type CacheEngine = "REDIS" | "KV" | "CACHE_API" | "CF_PROXY";
+export type CacheEngine =
+  | "REDIS"
+  | "KV"
+  | "CACHE_API"
+  | "FILE_SYSTEM"
+  | "S3";
+
 const cacheImplByEngine: Record<CacheEngine, CacheStorageOption> = {
   REDIS: {
     implementation: redisCache,
@@ -28,9 +40,13 @@ const cacheImplByEngine: Record<CacheEngine, CacheStorageOption> = {
     implementation: globalThis.caches,
     isAvailable: typeof globalThis.caches !== "undefined",
   },
-  CF_PROXY: {
-    implementation: cachesProxy,
-    isAvailable: true,
+  FILE_SYSTEM: {
+    implementation: cachesFileSystem,
+    isAvailable: isFileSystemAvailable,
+  },
+  S3: {
+    implementation: cachesS3,
+    isAvailable: isS3Available,
   },
 };
 
@@ -41,15 +57,17 @@ for (const [engine, cache] of Object.entries(cacheImplByEngine)) {
   };
 }
 
-const eligibleCacheImplementations = WEB_CACHE_ENGINES.map((engine) =>
-  cacheImplByEngine[engine]
-).filter((engine) => engine.isAvailable).map((engine) => engine.implementation);
+const eligibleCacheImplementations = WEB_CACHE_ENGINES
+  .map((engine) => cacheImplByEngine[engine])
+  .filter((engine) => engine?.isAvailable)
+  .map((engine) => engine.implementation);
 
-const getCacheStorage = (): CacheStorage => {
+const getCacheStorage = (): CacheStorage | undefined => {
   if (eligibleCacheImplementations.length === 0) {
     return cacheImplByEngine[DEFAULT_CACHE_ENGINE].implementation;
   }
-  return compose(...eligibleCacheImplementations);
+
+  return createTieredCache(...eligibleCacheImplementations);
 };
 
 export const caches = getCacheStorage();
--- a/runtime/caches/s3.ts
+++ b/runtime/caches/s3.ts
@@ -1,25 +1,27 @@
-import { logger, tracer } from "../../observability/otel/config.ts";
-import { meter } from "../../observability/otel/metrics.ts";
-import { ValueType } from "../../deps.ts";
-import {
-  assertCanBeCached,
-  assertNoOptions,
-  withCacheNamespace,
-} from "./common.ts";
-import {
-  DeleteObjectCommand,
-  GetObjectCommand,
-  PutObjectCommand,
-  S3Client,
-} from "https://esm.sh/@aws-sdk/client-s3@3.513.0";
 import {
   compress,
   decompress,
   init as initZstd,
 } from "https://denopkg.com/mcandeia/zstd-wasm@0.20.2/deno/zstd.ts";
+import {
+  DeleteObjectCommand,
+  GetObjectCommand,
+  NoSuchKey,
+  PutObjectCommand,
+  S3Client,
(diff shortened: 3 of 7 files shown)


runtime/caches/mod.ts
```diff
@@ -1,20 +1,34 @@
...
+  caches as cachesFileSystem,
+  isFileSystemAvailable,
+} from "./fileSystem.ts";
+import { caches as cachesS3, isS3Available } from "./s3.ts";
+import { createTieredCache } from "./tiered.ts";
 
-const DEFAULT_CACHE_ENGINE: CacheEngine = "CF_PROXY";
+const ENABLE_LOADER_CACHE = Deno.env.get("ENABLE_LOADER_CACHE") === "true";
+const DEFAULT_CACHE_ENGINE = ENABLE_LOADER_CACHE
+  ? "FILE_SYSTEM,S3"
```
ItamarRocha, line 14: Instead, shouldn't we set the default to KV? Since this is going to be released in the new deco version?

runtime/caches/fileSystem.ts
```diff
@@ -62,9 +62,9 @@ function createFileSystemCache(): CacheStorage {
       const fileContent = await Deno.readFile(filePath);
       return fileContent;
     } catch (err) {
-      if (err.code === "ENOENT") {
-        logger.warning(`file not found when reading from file system, path: ${FILE_SYSTEM_CACHE_DIRECTORY}/${key}`);
-      }else {
+      // Error code different for file/dir not found
+      // The file won't be found in cases where it's not cached
+      if (err.code !== "ENOENT") {
         logger.error(`error when reading from file system, ${err}`);
```
mcandeia: this should be removed right? the annoying message that we mentioned

runtime/caches/mod.ts
ItamarRocha, line 14: Instead, shouldn't we set the default to KV? Since this is going to be released in the new deco version?

runtime/caches/fileSystem.ts
```diff
@@ -240,6 +236,6 @@ const hasWritePerm = async (): Promise<boolean> => {
   ).then((status) => status.state === "granted");
 };
 
-export const caches = await hasWritePerm() && FILE_SYSTEM_CACHE_DIRECTORY
-  ? createFileSystemCache()
-  : undefined;
+export const isFileSystemAvailable = await hasWritePerm() &&
+  FILE_SYSTEM_CACHE_DIRECTORY !== undefined;
+export const caches = createFileSystemCache();
```
mcandeia, line 245: you can move this check to mod.ts instead

runtime/caches/mod.ts
```diff
@@ -46,10 +68,7 @@ const eligibleCacheImplementations = WEB_CACHE_ENGINES.map((engine) =>
 ).filter((engine) => engine.isAvailable).map((engine) => engine.implementation);
 
 const getCacheStorage = (): CacheStorage => {
-  if (eligibleCacheImplementations.length === 0) {
```
mcandeia, line 49: you should keep this default_cache_engine code as this is a breaking change if you have defined the variable but the value is invalid/empty

runtime/caches/s3.ts
```diff
@@ -78,18 +72,20 @@ function createS3Caches(): CacheStorage {
       accessKeyId: awsAccessKeyId,
       secretAccessKey: awsSecretAccessKey,
     },
-    useAccelerateEndpoint: true,
+    // useAccelerateEndpoint: true,
```
mcandeia, line 75: remove or comment why it is commented

runtime/caches/s3.ts
```diff
@@ -194,32 +190,28 @@ function createS3Caches(): CacheStorage {
...
               return undefined;
             }
-            span.addEvent("cache-hit");
 
-            const parsedData: Metadata = typeof data === "string"
-              ? JSON.parse(data)
-              : data;
-            parsedData.body.buffer = bufferToObject(parsedData.body.buffer);
+            const zstd = data[0] === 1;
+            const buffer = data.slice(1);
```
mcandeia, line 201: seems magic, can you add a comment why?

runtime/caches/s3.ts
```diff
@@ -303,8 +295,9 @@ function createS3Caches(): CacheStorage {
...
-const isEndpointSet = (bucketName && awsRegion) || awsEndpoint;
-const areCredentialsSet = awsAccessKeyId && awsSecretAccessKey;
-export const caches = (isEndpointSet && areCredentialsSet)
-  ? createS3Caches()
-  : undefined;
+const isEndpointSet = (bucketName !== undefined && awsRegion !== undefined) ||
+  awsEndpoint !== undefined;
+const areCredentialsSet = awsAccessKeyId !== undefined &&
+  awsSecretAccessKey !== undefined;
+export const isS3Available = isEndpointSet && areCredentialsSet;
```
mcandeia, line 302: move these checks to mod.ts

runtime/caches/tiered.ts
```diff
@@ -0,0 +1,141 @@
...
+const inFuture = (maybeDate: string) => {
+  try {
+    return new Date(maybeDate) > new Date();
+  } catch {
+    return false;
+  }
+};
+const isCache = (c: Cache | undefined): c is Cache => typeof c !== "undefined";
+
+export function createTieredCache(
```
mcandeia, line 12: Could you add a JSDoc of this function? just explaining the overall behavior

runtime/caches/tiered.ts
```diff
@@ -0,0 +1,141 @@
...
+    },
+    match: (
+      _request: URL | RequestInfo,
+      _options?: MultiCacheQueryOptions | undefined,
+    ): Promise<Response | undefined> => {
+      throw new Error("Not Implemented");
+    },
+    open: async (cacheName: string): Promise<Cache> => {
+      let maybeCache: Cache | undefined;
+      for (const caches of tieredCaches) {
```
mcandeia, line 45: filter it first, then open later

mcandeia:
```suggestion
      await Promise.all(tieredCaches.filter(Boolean).map(async cache => {
          openedCaches.push(await cache.open(cacheName));
      }))
```

mcandeia: replace the code below for the implementation that I've provided

runtime/caches/tiered.ts
```diff
@@ -0,0 +1,141 @@
...
+              maybeCache = undefined;
+            });
+          if (isCache(maybeCache)) {
+            openedCaches.push(maybeCache);
+          }
+        } else {
+          logger.error("No Cache available");
+        }
+      }
+      logger.info(`Tiered cache opened ${openedCaches.length} caches`);
```
mcandeia, line 60: not necessary

runtime/caches/tiered.ts
```diff
@@ -0,0 +1,141 @@
...
+            .then((c) => maybeCache = c)
+            .catch((error) => {
+              console.error("Error caught:", error);
+              maybeCache = undefined;
+            });
+          if (isCache(maybeCache)) {
+            openedCaches.push(maybeCache);
+          }
+        } else {
+          logger.error("No Cache available");
```
mcandeia, line 57: just noisy, I think

runtime/caches/tiered.ts
```diff
@@ -0,0 +1,141 @@
...
+
+            const expires = matched.headers.get("expires");
+            const isStale = expires ? !inFuture(expires) : false;
+
+            if (isStale) {
+              indexOfCachesToUpdate.push(index);
+            } else {
+              // found a match, no need to check the rest
+              break;
+            }
```
mcandeia, line 110: its not clear to me why are you considering that stale cache will be updated with the same value that returns from match ?

runtime/caches/tiered.ts
```diff
@@ -0,0 +1,141 @@
...
+          }
+          if (matched) {
+            updateTieredCaches(indexOfCachesToUpdate, request, matched);
+            return matched;
+          } else {
+            return undefined;
+          }
```
mcandeia, line 117: nit: just a suggestion
```suggestion
matched && updateTieredCaches(indexOfCachesToUpdate, request, matched);
return matched;
```

runtime/caches/tiered.ts
```diff
@@ -0,0 +1,141 @@
...
+        ): Promise<ReadonlyArray<Response>> => {
+          throw new Error("Not Implemented");
+        },
+        /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Cache/put) */
+        put: async (
+          request: RequestInfo | URL,
+          response: Response,
+        ): Promise<void> => {
+          const putPromises = openedCaches.map((caches) =>
+            caches.put(request, response.clone())
```
mcandeia, line 132: You have cloned inside `updateTieredCaches` method, so you're doing two body copies here, can you ensure that the body is copied only once per cache?

mcandeia (changes requested):

runtime/caches/fileSystem.ts
ItamarRocha: The

`
if (err.code !== "ENOENT")
`

Takes care of the annoying message, the error will only be called in case the error is not a *not found*/*not existing* (which was the annoying error).

runtime/caches/tiered.ts
ItamarRocha: Will do

runtime/caches/tiered.ts
ItamarRocha: If a match that has not expired is still available, we can use it instead of making the request again. Once all the matches are expired it will update with a new request response.

runtime/caches/mod.ts
ItamarRocha: I have changed it back to KV

runtime/caches/tiered.ts
ItamarRocha: It is only copied once per request, isn't it? The updateTieredCaches is called in a different context. The put inside the updateTieredCaches is not the same in the TieredCaches, it is the put from each cache inside the tiered cache.

runtime/caches/mod.ts
```diff
@@ -1,20 +1,53 @@
...
+  : DEFAULT_CACHE_ENGINE.split(",") as CacheEngine[];
+
+const FILE_SYSTEM_CACHE_DIRECTORY =
+  Deno.env.get("FILE_SYSTEM_CACHE_DIRECTORY") ?? undefined;
+
+const bucketName = Deno.env.get("CACHE_UPLOAD_BUCKET");
+const awsRegion = Deno.env.get("CACHE_AWS_REGION");
+const awsAccessKeyId = Deno.env.get("CACHE_AWS_ACCESS_KEY_ID")!;
+const awsSecretAccessKey = Deno.env.get("CACHE_AWS_SECRET_ACCESS_KEY")!;
+const awsEndpoint = Deno.env.get("CACHE_AWS_ENDPOINT");
```
mcandeia, line 22: this is duplicated from s3.ts file, you should either:


1) pass as a parameter when creating the s3 caches
2) put back inside the s3.ts file (sorry I think it might be an option to put back the previous approach I think its really better)

runtime/caches/tiered.ts
mcandeia: I didn't follow you, lets sync offline

runtime/caches/tiered.ts
mcandeia: that makes sense, can you check if the underlying caches implementation are avoiding clone since you're cloning it here?

runtime/caches/s3.ts
```diff
@@ -29,6 +30,7 @@ const awsRegion = Deno.env.get("CACHE_AWS_REGION");
 const awsAccessKeyId = Deno.env.get("CACHE_AWS_ACCESS_KEY_ID")!;
 const awsSecretAccessKey = Deno.env.get("CACHE_AWS_SECRET_ACCESS_KEY")!;
 const awsEndpoint = Deno.env.get("CACHE_AWS_ENDPOINT");
+const awsFolder = Deno.env.get("DECO_SITE_NAME") ?? "general_cache_deco";
```
mcandeia, line 33: Interesting. But we have multi-tenant sites. We should use `Context.active().site` instead.

runtime/caches/tiered.ts
```diff
@@ -0,0 +1,136 @@
...
+        /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Cache/delete) */
+        delete: async (
+          request: RequestInfo | URL,
+          options?: CacheQueryOptions,
+        ): Promise<boolean> => {
+          let deleteResponse;
+          for (const caches of openedCaches) {
+            deleteResponse = await caches.delete(request, options);
+          }
+          return deleteResponse ?? false;
```
mcandeia, line 75: do in parallel.

Promise.all(openedCaches.map(cache => cache.delete(request, options)).then(results => results.every(r => r))

mcandeia (approved):
fix then shipit

runtime/caches/tiered.ts
```diff
@@ -0,0 +1,134 @@
...
+    match: (
+      _request: URL | RequestInfo,
+      _options?: MultiCacheQueryOptions | undefined,
+    ): Promise<Response | undefined> => {
+      throw new Error("Not Implemented");
+    },
+    open: async (cacheName: string): Promise<Cache> => {
+      await Promise.all(
+        tieredCaches.filter(Boolean).map(async (cache) => {
+          if (cache) {
```
tlgimenes, line 52: why both if and filter?

tlgimenes (approved):

PR #608, opened 2024-03-11
Tavano/fix deco segment

guitavano:


runtime/fresh/middlewares/3_main.ts
```diff
@@ -182,35 +183,41 @@ export const handler = [
...
       for (const flag of state.flags) {
-        if (flag.value) {
-          active.add(flag.name);
-          inactiveDrawn.delete(flag.name);
-        } else {
-          active.delete(flag.name);
-          inactiveDrawn.add(flag.name);
+        // non-saved matcher includes ".rule" eg: "deco.matchers.1.rule"
+        // to avoid large cookies we only set saved matchers
+        const isSavedMatcher = !flag.name.includes(".rule");
```
mcandeia, line 197: This is a very weak check, try to do this instead:

1) create a boolean value that starts with true and set to false at the very first loop right here => https://github.com/deco-cx/deco/blob/main/blocks/matcher.ts#L143

example

```tsx

let firstClass = true;
// loop
....
 // stop on first resolvable
if (type === "resolvable") {
  break;
}
firstClass = false;
}
```

Because in case of resolvable is reached in the first loop so the matcher is "saved" (lets say firstClass instead)

mcandeia: Then, change the type flag to have a prop `firstClass?: boolean`

https://github.com/deco-cx/deco/blob/60d219ac0f5c6e5ede374c451a9f96818c41cc59/types.ts#L76

mcandeia: then, push it right here: https://github.com/deco-cx/deco/blob/60d219ac0f5c6e5ede374c451a9f96818c41cc59/blocks/matcher.ts#L172

and reuse elsewhere

utils/cookies.ts
```diff
@@ -0,0 +1,49 @@
...
+interface SetCookieOptions {
+  // Encoding a cookie is important to ensure special characters
+  // do not interfere with cookie parsing and to prevent the injection
+  // of malicious content, enhancing data security and integrity.
+  encode?: boolean;
+  // This option is used to bypass the 4KB cookie size limit.
+  dangerouslySetBigCookies?: boolean;
+}
+
+export function setCookie(
```
mcandeia, line 19: great!

utils/cookies.ts
```diff
@@ -0,0 +1,49 @@
+import { Cookie, setCookie as DenoSetCookie } from "std/http/cookie.ts";
```
mcandeia, line 1:
```suggestion
import { Cookie, setCookie as stdSetCookie } from "std/http/cookie.ts";
```

utils/cookies.ts
```diff
@@ -0,0 +1,49 @@
...
+): void {
+  const { encode = false, dangerouslySetBigCookies = false } = cookieOptions ??
+    {};
+
+  if (encode) {
+    cookie.value = btoa(encodeURIComponent(cookie.value));
+  }
+
+  // could have an error range, because deno's cookie uses their own toString function
+  const cookieString = JSON.stringify(cookie);
```
mcandeia, line 32: I think you should check `cooke.value+cookie.name` size instead. STRINGIFY is a very heavy operation

utils/cookies.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  if (!dangerouslySetBigCookies) {
+    const sizeInBytes = new TextEncoder().encode(cookieString).length;
```
mcandeia, line 35: theoretically you can just multiple the length * 2 and that's all.

mcandeia (changes requested):

utils/cookies.ts
```diff
@@ -0,0 +1,49 @@
...
+        `Cookie '${cookie.name}' exceeds the size limit of 4KB and will not be set.`,
+      );
+      return;
+    }
+  }
+
+  DenoSetCookie(headers, cookie);
+}
+
+export function decodeCookie(cookie: string): string {
```
mcandeia, line 47: should this be a `getCookie` by deco instead? that receive a { decode?: boolean } (same as you do in setCookie?)

mcandeia (changes requested):

utils/cookies.ts
guitavano: getCookie receive all header and does a for in each cookie, we can not use decode to all. Let's think about that in the future.

mcandeia (approved):

lucis (approved):

PR #611, opened 2024-03-13
remove resolvechain from cache key

hugo-ccabral:


mcandeia (approved):

mcandeia:
We can’t do that as the name of the loader is also part of the cache key

PR #613, opened 2024-03-13
fix invoke resolve chain

mcandeia:
It depends on fixing all usages of cache inside deco-cx/apps since its being used cacheBySegment and ignoring props.

Currently this is working because resolveChain is empty, hence the cache is being ignored when using invoke


engine/manifest/defaults.ts
```diff
@@ -181,10 +181,26 @@ export default {
...
+        }, {
+          type: "resolver",
+          value: __resolveType,
+        }],
+      });
     }
+    console.log(...ctx?.resolveChain ?? [], {
+      type: "resolver",
+      value: __resolveType,
+    })
```
ItamarRocha, line 197:
```suggestion
```

ItamarRocha (approved):

PR #646, opened 2024-04-09
change pathname to a generic admin route

guitavano:


guitavano:
Samething here: https://github.com/deco-cx/apps/pull/512 and approved

PR #696, opened 2024-06-03
Liveness probe

mcandeia:
# Pull Request Description

### Summary
This PR introduces a middleware that performs liveness checks on the application based on various metrics. The liveness checks can be configured via environment variables, and a `DRY_RUN` mode is available for testing purposes without causing actual failures.

### Key Features
1. **Metrics Interface**:
   - Defines the structure for the application's metrics, including uptime, request count, and average latency.

2. **LiveChecker Interface**:
   - Specifies the format for individual liveness checkers, which return a boolean value indicating the health of the service.

3. **Environment Variable Configuration**:
   - Probes can be configured through environment variables, allowing customization of thresholds for each checker.
   - The environment variable format is `PROBE_<CHECKER_NAME>_THRESHOLD`.

4. **DRY_RUN Mode**:
   - Introduces a `DRY_RUN` mode controlled by the `PROBE_DRY_RUN` environment variable.
   - When `DRY_RUN` is enabled, the system will not fail even if checks fail; instead, it will always return a status of 200 OK.

5. **Liveness Path**:
   - A liveness endpoint (`/_liveness`) is created, which performs all the configured checks and returns the appropriate HTTP status based on the results.

6. **Metrics Collection**:
   - Collects and updates metrics such as request count and average latency for use in liveness checks.

7. **Probe Metrics**:
   - Uses OpenTelemetry to create a counter metric (`failed_probe`) to keep track of failed liveness checks.

### Detailed Changes
- **Metrics Definition**:
  ```typescript
  export interface Metrics {
    uptime: number;
    requestCount: number;
    latency: {
      avg: number;
    };
  }
  ```

- **Liveness Checker Interface**:
  ```typescript
  export interface LiveChecker {
    name: string;
    checker: (metrics: Metrics) => Promise<boolean> | boolean;
  }
  ```

- **Environment Variable Handling**:
  ```typescript
  const envObj = Deno.env.toObject();
  const DRY_RUN = envObj.PROBE_DRY_RUN === "true";
  export const getProbeThresholdAsNum = (checkerName: string): number | undefined => {
    const fromEnv = envObj[`PROBE_${checkerName}_THRESHOLD`];
    return fromEnv ? +fromEnv : undefined;
  };
  ```

- **Liveness Middleware Handler**:
  ```typescript
  const buildHandler = (...checkers: LiveChecker[]): MiddlewareHandler => {
    // Metrics and request handling logic
  };
  
  export const liveness = buildHandler(
    memoryChecker,
    upTimeChecker,
    reqCountChecker,
    avgLatencyChecker,
  );
  ```

### Usage
- Set environment variables to configure probe thresholds, e.g., `PROBE_MEMORY_THRESHOLD=80`.
- Enable DRY_RUN mode for testing: `PROBE_DRY_RUN=true`.
- Access the liveness endpoint at `/_liveness` to check the application's health status based on the configured probes.


observability/probes/handler.ts
```diff
@@ -0,0 +1,95 @@
...
+          name: failedCheck.name,
+        });
+        return new Response(checks, { status });
+      }
+      return new Response(checks, { status: 200 });
+    }
+    reqCount++;
+    const start = performance.now();
+    return ctx.next().finally(() => {
+      avgLatency = (avgLatency * (reqCount - 1) + performance.now() - start) /
```
hugo-ccabral, line 84: You are incrementing forever, we have a few cases that have stable deployment for long periods, this could drive to overflows. What about use a sliding window of 500 request or so?

observability/probes/handler.ts
```diff
@@ -0,0 +1,95 @@
...
+        reqCount;
+    });
```
matheusgr, line 86: maybe we can count requests at two moments:

- when a request start (reqCount)
- when it finishes (reqFinishCount)

observability/probes/memory.ts
```diff
@@ -0,0 +1,18 @@
+import { getProbeThresholdAsNum, type LiveChecker } from "./handler.ts";
+
+const HEALTHY_RATIO = Deno.env.get("PROBE_MEMORY_HEALTHY_RATIO");
+
+const healthyRatio = HEALTHY_RATIO ? parseFloat(HEALTHY_RATIO) : 0.81;
```
matheusgr, line 5: Why should we care about HEALTHY_RATIO?

It seems a clumsy or unnecessary metric. RSS max total should what we care.

matheusgr (changes requested):

observability/probes/memory.ts
```diff
@@ -0,0 +1,18 @@
...
+
+const NAME = "MEMORY";
+const MAX_MEM_THRESHOLD_MB = getProbeThresholdAsNum(NAME);
+
+export const memoryChecker: LiveChecker = {
+  name: NAME,
+  checker: () => {
+    const available = Deno.systemMemoryInfo().available / 1024;
```
hugo-ccabral, line 13: What does this return? The resource limit from the pod or the node?

observability/probes/uptime.ts
```diff
@@ -0,0 +1,18 @@
+import {
+  getProbeThresholdAsNum,
+  type LiveChecker,
+  type Metrics,
+} from "./handler.ts";
+
+const NAME = "UPTIME";
+const MAX_REQ_THRESHOLD = getProbeThresholdAsNum(NAME);
```
hugo-ccabral, line 8:
```suggestion
const MAX_UPTIME_THRESHOLD = getProbeThresholdAsNum(NAME);
```

observability/probes/handler.ts
```diff
@@ -0,0 +1,95 @@
...
+  };
+};
+
+export const liveness = buildHandler(
```
hugo-ccabral, line 90: buildHandler is not exported

runtime/fresh/middlewares/0_liveness.ts
```diff
@@ -0,0 +1,4 @@
+import { buildHandler } from "../../../observability/probes/handler.ts";
+import { memoryChecker } from "../../../observability/probes/memory.ts";
+
+export const liveness = buildHandler(memoryChecker);
```
hugo-ccabral, line 4: buildHandler is not exported

hugo-ccabral (approved):

hugo-ccabral (changes requested):
Sliding window is the most important change I guess, also double check the buildHandler if it is exported

observability/probes/avgLatency.ts
```diff
@@ -0,0 +1,18 @@
+import {
+  getProbeThresholdAsNum,
+  type LiveChecker,
+  type Metrics,
+} from "./handler.ts";
+
+const NAME = "AVG_LATENCY";
+const AVG_LATENCY_THRESHOLD = getProbeThresholdAsNum(NAME);
+
+export const avgLatencyChecker: LiveChecker = {
```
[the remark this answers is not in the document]
mcandeia: Agree, I think it is a different probe, basically, but avg latency still valid as is

observability/probes/handler.ts
```diff
@@ -0,0 +1,95 @@
...
+    ) {
+      const results = await Promise.all(
+        checkers.map(async ({ checker, name }) => {
+          return { check: await checker(metrics), name };
+        }),
+      );
+      const failedCheck = results.find(({ check }) => !check);
+      const checks = JSON.stringify(results, null, 2);
+      if (failedCheck) {
+        const status = DRY_RUN ? 200 : 503;
```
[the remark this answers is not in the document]
mcandeia: this is configurable through liveness probe configurations in pod

observability/probes/handler.ts
mcandeia: inflight requests you mean? that's a different probe, I can do that

observability/probes/memory.ts
mcandeia: But, if the memory reaches the RSS so the pod will get killed by k8s anyways, how it should be then?

observability/probes/memory.ts
mcandeia: node

mcandeia: but it was removed.

matheusgr (approved):

PR #723, opened 2024-07-25
Long live to Hono!

mcandeia:


tlgimenes:
I feel like we have a third phase. The third phase encompass moving all features of htmx related framework into htmx and removing from deco.

Features like Async Rendering should be only on the htmx framework for instance. Loaders cache could be shared across frameworks.

tlgimenes (approved):

guifromrio:
Imagine doing this before TypeScript. And some people insist TS is useless. LOL!

Amazing work!

igorbrasileiro:
@mcandeia what do you think about handling responder.headers changes that loaders can do? ctx.response.headers.set('key', 'val').

This would be useful to use early hints 103 headers

mcandeia:
> @mcandeia what do you think about handling responder.headers changes that loaders can do at htmx renderFn? ctx.response.headers.set('key', 'val').
> 
> This would be useful to use early hints 103 headers

AFAIK response.headers is used when building the final response

PR #724, opened 2024-07-29
Add TLD on CSP frame ancestor

matheusgr:


utils/admin.ts
```diff
@@ -5,7 +5,7 @@ export const adminDomains = [
   "https://v0-admin.deco.cx/",
   "https://play.deco.cx/",
 ];
-export const landingPageDomain = `https://www.deco.cx`;
+export const landingPageDomain = ["https://www.deco.cx", "https://www.deco.cx"];
```
igorbrasileiro, line 8: Does this second URL should be the same as it already exists?

igorbrasileiro (approved):

igorbrasileiro:
In the future, these headers could be added via environment variables, decoupling from deco knows about the admin URL and allowing users to add their frame-ancestors. Does it make sense?

igorbrasileiro (approved):

PR #733, opened 2024-08-01
first class multipart support

viktormarinho:
FormData support, where you never have to touch a FormData object :)

usage:

```ts
// Client-side
const file = new File(["a"], "foo.txt");

await invoke.actions.upload({ file }, { multipart: true });

// Server-side
export interface Props {
  file: File
}

export async function upload(props: Props, _req: Request, _ctx: AppContext) {
  console.log(props.file.name) // "foo.txt"!
}
```


viktormarinho:
Future improvement: make it into a custom protocol, probably adding field metadata into the formdata keys, so that i can handle serializing/deserializing booleans/numbers automatically as well.

Current implementation just uses FormData so everything gets converted to string or file. I also think thats okay for this first version, people just need to know that setting `multipart` to true have this edge case

viktormarinho:
One thing that this is doing that should be addressed as well: We're always serializing/deserializing the props into formdata on invokes with multipart: true. This should not happen in server-side since we dont need to pass the props through http

clients/formdata_test.ts
matheusgr, line 1: we use `.test.ts` extension for tests.

clients/formdata.ts
```diff
@@ -0,0 +1,91 @@
...
+export function propsToFormData(props: unknown): FormData {
+  if (props instanceof FormData) {
+    return props;
+  }
+
+  const formData = new FormData();
+
+  const appendToFormData = (
+    _key: string,
+    value: unknown,
```
matheusgr, line 18: since we don't have conversion type yep, maybe require value as string | object | undefined, and avoid other types.

clients/formdata.ts
```diff
@@ -0,0 +1,91 @@
...
+                String(nestedValue),
+              );
+            }
+          }
+        } else {
+          if (item instanceof Blob) {
+            formData.append(`${key}.${index}`, item);
+          } else {
+            formData.append(`${key}.${index}`, String(item));
+          }
```
matheusgr, line 41: if undefined, don't add the index.

clients/withManifest.ts
```diff
@@ -92,13 +94,29 @@ const fetchWithProps = async (
...
 
-  const response = await (init?.fetcher ?? fetch)(url, {
+  if (!headers.has("content-type")) {
+    headers.set("content-type", "application/json");
+  }
+
+  if (multipart) {
+    // Let the browser handle the content-type header, with a webkit generated boundary
+    headers.delete("content-type");
+  }
```
matheusgr, line 109: it seems that an error should be thrown here. If there is set a content-type, but also you want multipart.. maybe you did something wrong

matheusgr (changes requested):

runtime/fresh/routes/invoke.ts
```diff
@@ -1,11 +1,41 @@
...
+  req: Request,
+): Promise<Record<string, any>> {
+  const contentType = req.headers.get("content-type");
+  const contentLength = req.headers.get("content-length");
+
+  if (contentLength === "0") {
+    return {};
+  }
+
+  if (contentType === "application/json") {
```
mcandeia, line 28: API-wise (not considering the client that is making the call) this is a breaking change, right? you can use `"Content-Type":"application/json; charset=utf-8",` (which is very common) and then it will be parsed as `{}` , should it defaults to json instead?

I think the best way is to have one `content-type` parser which returns the actual content-type of the given body and a map from `content-type` to `bodyParser`. Otherwise fallsback to json() or {} (if content-length is not present or zero)

clients/withManifest.ts
mcandeia: Agree, is not much intuitive deleting a content-type header, it should not be added or let the error occur

clients/formdata.ts
```diff
@@ -0,0 +1,91 @@
+function escapeDots(key: string): string {
```
mcandeia, line 1: I think it might worth to add some Jsdoc since these functions will be widely used

mcandeia (changes requested):

clients/withManifest.ts
viktormarinho: Which one do you guys think its better:
- do a console.warn that your content-type header is being overriden since you're using multipart: true
- throw an error if you try to set both

viktormarinho: @mcandeia @matheusgr

clients/withManifest.ts
```diff
@@ -92,13 +94,35 @@ const fetchWithProps = async (
...
-  headers.set("content-type", "application/json");
 
-  const response = await (init?.fetcher ?? fetch)(url, {
+  if (multipart && headers.has("content-type")) {
+    // throw new Error(
+    //   "Setting the content-type header is not allowed when using multipart",
+    // );
+
+    console.warn(
+      "[Deco] Invoke: Setting the content-type header is not allowed when using multipart, it will be ignored",
```
mcandeia, line 108:
```suggestion
      `invoke ${url}: Setting the content-type header is not allowed when using multipart, it will be ignored`,
```

mcandeia (approved):

PR #740, opened 2024-08-02
stop throw without ctx

guitavano:


viktormarinho (approved):
For me, looks good. Maybe we should ask some more people to see if they have a problem with this.

PR #750, opened 2024-08-06
[Hono fix] Set response to undefined to force not deleting headers

mcandeia:
For some (akward) reason, hono deletes `content-type` if set response is not fresh (meaning everytime that any middleware runs). This leads in `headers are immutable` error when certain conditions like redirects or websocket connections. So i'm first setting it to undefined to make sure response is always fresh.


viktormarinho (approved):
looks ok

PR #754, opened 2024-08-07
Fix/middleware globals

viktormarinho:
Fix: Middleware globals
Fix: isEventStreamResponse being exported from server only module

### DIFF
--- a/runtime/fresh/plugin.tsx
+++ b/runtime/fresh/plugin.tsx
@@ -9,7 +9,27 @@ import framework from "./Bindings.tsx";
 export interface Plugin {
   name: string;
   routes?: PluginRoute[];
+  middlewares?: PluginMiddleware[];
 }
+
+export interface PluginMiddleware {
+  path: string;
+  middleware: Middleware;
+}
+
+export interface Middleware {
+  handler: MiddlewareHandler | MiddlewareHandler[];
+}
+
+export type MiddlewareHandler = (
+  req: Request,
+  ctx: {
+    next: () => Promise<Response | undefined>;
+    // deno-lint-ignore no-explicit-any
+    state: any;
+    params: Record<string, string>;
+  },
+) => Promise<Response | undefined>;
 
 export interface PluginRoute {
   /** A path in the format of a filename path without filetype */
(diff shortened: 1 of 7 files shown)


runtime/fresh/plugin.tsx
```diff
@@ -39,31 +64,61 @@ export type OptionsProvider<TManifest extends AppManifest = AppManifest> = (
...
-export default function decoPlugin(opt: Options): Plugin {
+export const component = ({ data }: PageParams<PageData>) => {
+  return <data.page.Component {...data.page.props} />;
+};
+
+export function createFreshHandler(deco: Deco) {
+  const h: PluginRoute["handler"] = (req: Request, ctx) => {
+    return deco.handler(req, {
+      RENDER_FN: ctx.render.bind(ctx),
+      GLOBALS: ctx.state,
```
mcandeia, line 75:
```suggestion
      GLOBALS: ctx.state.global,
```

runtime/fresh/plugin.tsx
```diff
@@ -39,31 +64,61 @@ export type OptionsProvider<TManifest extends AppManifest = AppManifest> = (
...
+              await deco.prepareState({
+                req: { raw: req, param: () => ctx.params },
+              }),
+            );
+
+            return await ctx.next();
+          },
+        },
+      },
+    ],
```
mcandeia, line 121: remove this as this should be optional

runtime/fresh/plugin.tsx
```diff
@@ -39,31 +64,61 @@ export type OptionsProvider<TManifest extends AppManifest = AppManifest> = (
...
-    },
-    handler: async (req: Request, ctx) => {
-      const hdnl = await handlerPromise;
-      return hdnl(req, {
-        RENDER_FN: ctx.render.bind(ctx),
-      });
-    },
+    component,
+    handler: (req: Request, ctx) =>
+      createFreshHandler(ctx.state.deco)(req, ctx),
```
mcandeia, line 99: do not assume that `ctx.state.deco` will be there, instead, await deco again.

mcandeia (changes requested):

PR #772, opened 2024-08-15
feat: check tailwind

devartes:
### **What does this do?**
The script helps people find dynamic classes in the code more quickly. It will generate a table with the classes and you can click on the file name to go directly to the Tailwind class.
Captura de Tela 2024-08-15 às 08 46 28
To test this script, just put this version in the deco of deno.json, for example:
`"deco/": "https://cdn.jsdelivr.net/gh/deco-cx/deco@dbd9bb1a68b4830f267f8f66736ceaebd5e5ba7b/"`, and also in deno.json where it says "tasks", put something like:
`"tailwind-checker": "deno eval 'import \"deco/scripts/tailwind_checker.ts\"'"` , the name "tailwind-checker" is not mandatory, you can choose another name and run it as `deno task nameyouchose` : `"nameyouchose": "deno eval 'import \"deco/scripts/tailwind_checker.ts\"'"`

about this part of the code:
```
const green = "\x1b[32m";
const reset = "\x1b[0m";
```

https://talyian.github.io/ansicolors/
https://github.com/shiena/ansicolor/blob/master/README.md


scripts/tailwind_checker.ts
```diff
@@ -0,0 +1,131 @@
+import { walk } from "https://deno.land/std@0.201.0/fs/mod.ts";
```
mcandeia, line 1: please, change to `@std/fs/walk` so we can use JSR specifiers

scripts/tailwind_checker.ts
```diff
@@ -0,0 +1,131 @@
...
+
+const green = "\x1b[32m";
+const reset = "\x1b[0m";
+
+const folderPaths = ["./components", "./sections", "./islands"];
```
mcandeia, line 6: do we need to filter per folder? can't we just search for all `.tsx` except the ones that is listed at `.gitignore` (like node_modules)?

scripts/tailwind_checker.ts
```diff
@@ -0,0 +1,131 @@
...
+      }
+    } else {
+      console.log("No classes found in CSS or project files.");
+    }
+  } catch (error) {
+    console.error(`Error comparing classes: ${error}`);
+  }
+}
+
+compareClasses();
```
mcandeia, line 131: important as this will run whenever the script is included in the bundle using JSR (even when you don't want to run it)
```suggestion
if (import.meta.main) {
    await compareClasses();
}
```

scripts/tailwind_checker.ts
```diff
@@ -0,0 +1,131 @@
+import { walk } from "https://deno.land/std@0.201.0/fs/mod.ts";
+
+const green = "\x1b[32m";
+const reset = "\x1b[0m";
```
mcandeia, line 4: nit: you can use `@std/fmt/colors`

scripts/tailwind_checker.ts
```diff
@@ -0,0 +1,131 @@
...
+
+const folderPaths = ["./components", "./sections", "./islands"];
+const tailwindCssUrl =
```
mcandeia, line 7: feeling that it should be get from somewhere related to the project's tailwind version or at least a way to specify as an env var, but you can keep it here

mcandeia (changes requested):
You need to include your script here:

https://github.com/deco-cx/deco/blob/main/scripts/deno.json#L7

And when published you can run as `deno run -A jsr:@deco/scripts/tailwind-checker`

mcandeia:
also, I think you need to change the case to kebab_case to camelCase

mcandeia:
great job btw!

scripts/tailwind_checker.ts
devartes: About the env, ok about not doing this now.

PR #784, opened 2024-08-20
fix(loader): prevent sections not being cached when ENABLE_LOADER_CACHE is false

igorbrasileiro:
This pr enables sites that uses websites/sections/Rendering/Lazy.tsx to cache their responses even if ENABLE_LOADER_CACHE is false.

### DIFF
--- a/blocks/loader.ts
+++ b/blocks/loader.ts
@@ -180,12 +180,12 @@ const wrapLoader = (
       const isCacheModeToBypass = mode === "no-store";
 
       const bypassCache = isCacheModeToBypass ||
-        !isCacheEngineDefined ||
         isCacheKeyValueToBypass;
       try {
         // Should skip cache
         if (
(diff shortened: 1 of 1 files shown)


blocks/loader.ts
```diff
@@ -174,15 +174,23 @@ const wrapLoader = (
       const start = performance.now();
       let status: "bypass" | "miss" | "stale" | "hit" | undefined;
       const cacheKeyValue = cacheKey(props, req, ctx);
+      const bypassCache = mode === "no-store" ||
+        !isCache(maybeCache) ||
+        cacheKeyValue === null;
```
lucis, line 179: Can you split this line into different `const` and give them a explaining name?

lucis (approved):

blocks/loader.ts
igorbrasileiro: Sure, fixed

PR #802, opened 2024-08-28
add xml viewer

guitavano:
Screenshot 2024-08-28 at 09 27 37

### DIFF
--- a/engine/manifest/defaults.ts
+++ b/engine/manifest/defaults.ts
@@ -153,8 +153,10 @@ export default {
     const resolved = await resolve({ __resolveType: block, ...props }, {
       propsAreResolved: true,
     });
+    const isXml = resolved.headers.get("content-type").includes("xml");
     const preview = await resolve({
       __resolveType: pvResolver,
+      ...(isXml ? { __decoXml: await resolved.text() } : {}),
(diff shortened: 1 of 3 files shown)


engine/manifest/defaults.ts
```diff
@@ -153,8 +153,10 @@ export default {
     const resolved = await resolve({ __resolveType: block, ...props }, {
       propsAreResolved: true,
     });
+    const isXml = resolved.headers.get("content-type").includes("xml");
     const preview = await resolve({
       __resolveType: pvResolver,
+      ...(isXml ? { __decoXml: await resolved.text() } : {}),
       ...resolved,
     }, { propsAreResolved: true });
     return preview;
```
mcandeia, line 162:
```suggestion
    const resolved = await resolve({ __resolveType: block, ...props }, {
      propsAreResolved: true,
    });
    const preview = await resolve(new Proxy(resolved, { get: (val: any, prop: string, recv: any) => prop === "__resolveType" ? pvResolver: Reflect.get(val, prop, recv),
 }), { propsAreResolved: true });
    return preview;
```

mcandeia (changes requested):

PR #812, opened 2024-08-31
Make idle run when deps failed

mcandeia:


mcandeia:
approved offline by @hugo-ccabral

Issue #790, opened 2024-08-21
Fix typescript slow-types


mcandeia:
# Description

When publishing into JSR we are ignoring slow-types by adding `--allow-slow-types` [read more](https://jsr.io/docs/about-slow-types). Slow types should be removed in order to have better LSP.

To check where are the slow types occurrences just run `deno publish --dry-run --allow-dirty` and check the output.

Samuel-Brito19:
I'm working on a PR for this issue.

mcandeia:
thanks @Samuel-Brito19

PR #813, opened 2024-08-31
Adding return types to avoid compilation errors

Samuel-Brito19:
Issue #790 

In order to deal with the slow types errors,attempted to address them by adding explicit type annotations into the files where the problem occurs. 

I tried this approach and resolved 38 out of 41 issues. However, three errors persist. We can solve that in another PR.

### DIFF
--- a/clients/withManifest.ts
+++ b/clients/withManifest.ts
@@ -315,7 +315,7 @@ export const create = <TManifest extends AppManifest>() =>
 ): Promise<InvokeResult<TPayload, TManifest>> =>
   invokeKey(key, props, init) as Promise<InvokeResult<TPayload, TManifest>>;
 
-interface ManifestInvoke<TManifest extends AppManifest> {
+export interface ManifestInvoke<TManifest extends AppManifest> {
   invoke: ReturnType<typeof invoke<TManifest>>;
   create: ReturnType<typeof create<TManifest>>;
 }
@@ -368,14 +368,10 @@ export const proxy = <TManifest extends AppManifest>(
   return proxyFor(invoke<TManifest>(fetcher) as typeof batchInvoke);
 };
 
-interface forAppType<TApp extends App> {
-  create: ReturnType<typeof create<ManifestOf<TApp>>>;
-  invoke: InvocationProxyWithBatcher<ManifestOf<TApp>>;
-}
 /**
  * Creates a proxy that lets you invoke functions based on the declared actions and loaders. (compatibility with old invoke)
  */
-export const forApp = <TApp extends App>(): forAppType<TApp> => {
+export const forApp = <TApp extends App>(): ManifestInvoke<ManifestOf<TApp>> => {
   const { create } = withManifest<ManifestOf<TApp>>();
   return {
     create: create,
--- a/clients/withManifest.ts
+++ b/clients/withManifest.ts
@@ -315,15 +315,23 @@ export const create = <TManifest extends AppManifest>() =>
 ): Promise<InvokeResult<TPayload, TManifest>> =>
   invokeKey(key, props, init) as Promise<InvokeResult<TPayload, TManifest>>;
 
+interface WithManifestReturn<TManifest extends AppManifest> {
+  invoke: ReturnType<typeof invoke<TManifest>>;
+  create: ReturnType<typeof create<TManifest>>;
+}
+
 /**
  * Creates a set of strongly-typed utilities to be used across the repositories where pointing to an existing function is supported.
  */
-export const withManifest = <TManifest extends AppManifest>() => {
+export const withManifest = <
+  TManifest extends AppManifest,
+>(): WithManifestReturn<TManifest> => {
   return {
     /**
      * Invokes the target function using the invoke api.
      */
     invoke: invoke<TManifest>(),
+
     /**
      * Creates an invoker function. Usage:
      *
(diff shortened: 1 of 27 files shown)


observability/otel/metrics.ts
```diff
@@ -1,3 +1,4 @@
+import type { Meter } from "npm:@opentelemetry/api@1.9.0";
```
mcandeia, line 1: you shouldn't import from npm directly, import from `deps.ts`

engine/core/resolver.ts
```diff
@@ -107,7 +107,7 @@ const SCORE_BY_RESOLVE_TYPES = new Map(
 );
 
 export const FieldResolver = {
-  minify: (chain: FieldResolver[]) => {
+  minify: (chain: FieldResolver[]): (string | number | undefined)[] => {
```
mcandeia, line 110: Not sure if number/undefined is expected I think it should be somehow string[]

mcandeia (changes requested):
Great job, we are almost there, just few more improvements

runtime/mod.ts
```diff
@@ -130,13 +148,13 @@ export class Deco<TAppManifest extends AppManifest = AppManifest> {
...
 
-  invoke(...args: Parameters<typeof invoke>) {
+  invoke(...args: Parameters<typeof invoke>): Promise<any> {
     return invoke(...args);
   }
 
   batchInvoke(
     ...args: Parameters<typeof batchInvoke>
-  ) {
+  ): Promise<any> {
```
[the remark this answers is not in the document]
Samuel-Brito19: Understandable.

runtime/mod.ts
```diff
@@ -112,7 +126,7 @@ export class Deco<TAppManifest extends AppManifest = AppManifest> {
     previewUrl: string,
     props: unknown,
     ctx?: State<TAppManifest>,
-  ) {
+  ): Promise<PreactComponent<any>> {
```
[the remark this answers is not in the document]
Samuel-Brito19: Got it.

observability/otel/metrics.ts
Samuel-Brito19: I tried to use MeterProvider from deps.ts, but got this error:

`Type 'Meter' is missing the following properties from type 'MeterProvider': _sharedState, _shutdown, getMeter, addMetricReader, and 2 more.`

runtime/features/invoke.ts
```diff
@@ -71,7 +71,7 @@ export const invoke = async (
   props: InvokeFunction["props"],
   select: InvokeFunction["select"],
   state: State<any>,
-) => {
+): Promise<any> => {
```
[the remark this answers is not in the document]
Samuel-Brito19: The only step to see it is to initialize the storefront or I need to do something else?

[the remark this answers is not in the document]
mcandeia: 1. fork storefront at the same root folder as deco (/deco and /storefront at same dir level)
2. open storefront's `deno.json`
3. change `deco/` imports entry to `../deco/`
4. run `deno task check`

utils/timings.ts
```diff
@@ -5,7 +5,21 @@ export interface Timing {
   end?: number;
 }
 
-export function createServerTimings() {
+export interface ServerTimingsReturn {
```
mcandeia, line 8: the word `return` is redundant as everything can be "returned" from somewhere
```suggestion
export interface ServerTimingsBuilder {
```

clients/withManifest.ts
```diff
@@ -315,15 +315,23 @@ export const create = <TManifest extends AppManifest>() =>
 ): Promise<InvokeResult<TPayload, TManifest>> =>
   invokeKey(key, props, init) as Promise<InvokeResult<TPayload, TManifest>>;
 
+interface WithManifestReturn<TManifest extends AppManifest> {
```
mcandeia, line 318:
```suggestion
interface ManifestInvoke<TManifest extends AppManifest> {
```

clients/withManifest.ts
```diff
@@ -360,10 +368,14 @@ export const proxy = <TManifest extends AppManifest>(
   return proxyFor(invoke<TManifest>(fetcher) as typeof batchInvoke);
 };
 
+interface forAppType<TApp extends App> {
```
mcandeia, line 371: Is this type necessary? Isn't the same as `withManifest` ?

mcandeia (changes requested):

observability/otel/metrics.ts
mcandeia: Because there are more than one definition of meter, you need to import it correctly from opentelemetry/api.

clients/withManifest.ts
Samuel-Brito19: I think the difference is the `InvocationProxyWithBatcher` that I put in order to satisfy the `proxy` return type.

clients/withManifest.ts
```diff
@@ -315,15 +315,23 @@ export const create = <TManifest extends AppManifest>() =>
 ): Promise<InvokeResult<TPayload, TManifest>> =>
   invokeKey(key, props, init) as Promise<InvokeResult<TPayload, TManifest>>;
 
+interface ManifestInvoke<TManifest extends AppManifest> {
```
mcandeia, line 318: export
```suggestion
export interface ManifestInvoke<TManifest extends AppManifest> {
```

clients/withManifest.ts
mcandeia: I don't think so, I just verified that you can use it

clients/withManifest.ts
```diff
@@ -360,10 +368,14 @@ export const proxy = <TManifest extends AppManifest>(
...
 
+interface forAppType<TApp extends App> {
+  create: ReturnType<typeof create<ManifestOf<TApp>>>;
+  invoke: InvocationProxyWithBatcher<ManifestOf<TApp>>;
+}
 /**
  * Creates a proxy that lets you invoke functions based on the declared actions and loaders. (compatibility with old invoke)
  */
-export const forApp = <TApp extends App>() => {
+export const forApp = <TApp extends App>(): forAppType<TApp> => {
```
mcandeia, line 378:
```suggestion
export const forApp = <TApp extends App>(): : ManifestInvoke<
  ManifestOf<TApp>
> => {
```

mcandeia (changes requested):

PR #838, opened 2024-09-13
fix: burst props with href when href is set by user

igorbrasileiro:


igorbrasileiro:
TODO: add error message when running in localhost

hooks/useSection.ts
```diff
@@ -40,12 +77,16 @@ export const useSection = <P>(
 
   const { request, renderSalt, context: { state: { pathTemplate } } } = ctx;
 
+  const shouldBurstCache = (href?: string): href is string => href !== undefined && Object.keys(props).length === 0
```
igorbrasileiro, line 80: @guitavano give more context about `Object.keys(props).length ===0`

hooks/useSection.ts
```diff
@@ -40,12 +77,16 @@ export const useSection = <P>(
...
+
   const params = new URLSearchParams([
-    ["props", JSON.stringify(props)],
+    ["props",
+      // This is an workaround to vary the cache when useSection is used without props
+      JSON.stringify(shouldBurstCache(href) ? createPropsFromHref(href!) : props)],
```
mcandeia, line 85: wouldn't be better to vary cb instead?

mcandeia (changes requested):

mcandeia (approved):

PR #858, opened 2024-10-07
Replace the old redis cache implementation

azisaka:
I created a new version of the Redis cache engine.

Currently it supports the put, match and delete methods. The connection and methods will not block the load of the pages.

It was made to also support reconnection on failure; and even during a failure the website will keep running until the connection gets healthy again.

When an error occurs, the connection will be halted on purpose. Just like a open circuit breaker. The idea is to not fall into a reconnection loop. Once the website tries to cache again, the engine will try to reconnect.


runtime/caches/redis.ts
```diff
@@ -1,36 +1,28 @@
...
-    enableTelemetry: true,
-  })
-  : null;
-
-interface ResponseMetadata {
-  body: string;
-  status: number;
-  headers: [string, string][];
-}
+const REDIS_URL = Deno.env.get("REDIS_URL") || "redis://localhost:6379/0";
```
mcandeia, line 8: mistyped or expected?
```suggestion
const REDIS_URL = Deno.env.get("REDIS_URL") || "redis://localhost:6379/";
```

runtime/caches/redis.ts
```diff
@@ -1,36 +1,28 @@
...
+  );
 }
 
-function base64decode(str: string): string {
-  return decodeURIComponent(atob(str));
+function deserialize(raw: string): Promise<Response> {
+  return Promise.resolve(raw).then(JSON.parse).then((data) =>
+    new Response(data.body, { headers: data.headers, status: data.status })
+  );
 }
```
mcandeia, line 24:
```suggestion
function deserialize(raw: string): Response {
  const { body, headers, status } = JSON.parse(raw);
  return new Response(body, { headers, status })
}
```

runtime/caches/redis.ts
```diff
@@ -47,13 +39,33 @@ export const caches: CacheStorage = {
...
+      console.error(err);
+
+      redis.disconnect();
+    }).on("connect", () => {
+      console.debug("[cache]", "connecting");
+    }).on("reconnecting", () => {
+      console.warn("[cache]", "reconnecting");
+    });
+
+    const ensureConnection = () => {
```
mcandeia, line 55: ensure is returning boolean but I couldn't see any usages, can it just return void?

runtime/caches/redis.ts
```diff
@@ -47,13 +39,33 @@ export const caches: CacheStorage = {
...
+      if (!redis.isOpen && !redis.isReady) {
+        console.log("[cache]", "not ready");
+
+        redis.connect();
+
+        return false;
+      }
+
+      return true;
```
mcandeia, line 64: can connect be called multiple times? what would happen?

in case of both redis.isOpen and redis.isReady being false a sync connect() call will make those true? or it will be async set? if the 2nd option is true so you might have multiples `connect` calls

```suggestion
      if (!redis.isOpen && !redis.isReady) {
        console.log("[cache]", "not ready");

        redis.connect();

        return false;
      }

      return true;
```

runtime/caches/redis.ts
```diff
@@ -81,125 +89,78 @@ export const caches: CacheStorage = {
       ): Promise<ReadonlyArray<Request>> => {
         throw new Error("Not Implemented");
       },
-      /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Cache/match) */
+      delete: async (
+        request: RequestInfo | URL,
+        _?: CacheQueryOptions,
+      ): Promise<boolean> => {
+        console.debug("[cache]", "delete");
```
mcandeia, line 96: these methods are invoked multiples times per minute, does it worth to add logs in each call?

mcandeia: we can instead extract only the necessary info to keep it tracked? (and use a metric instead of a log?)

runtime/caches/redis.ts
```diff
@@ -81,125 +89,78 @@ export const caches: CacheStorage = {
...
+
+        ensureConnection();
+
+        return requestURLSHA1(request)
+          .then(redis.del)
+          .then(() => true)
+          .catch((err: Error) => {
+            console.error(err);
```
mcandeia, line 104: I think errors can be sent to hyperdx, import { logger } from "../../observability/mod.ts"

runtime/caches/redis.ts
```diff
@@ -81,125 +89,78 @@ export const caches: CacheStorage = {
...
+
+            return false;
+          });
+      },
       match: async (
         request: RequestInfo | URL,
         options?: CacheQueryOptions,
       ): Promise<Response | undefined> => {
+        console.debug("[cache]", "match");
```
mcandeia, line 113: remove

runtime/caches/redis.ts
```diff
@@ -47,13 +39,33 @@ export const caches: CacheStorage = {
...
-      throw new Error(
-        "Redis coult not be used due to the lack of credentials.",
-      );
-    }
-    const requestURLSHA1 = withCacheNamespace(cacheName);
+  open: async (_: string): Promise<Cache> => {
+    const redis = createClient({
+      url: REDIS_URL,
+    }).on("error", (err: Error) => {
+      console.error(err);
```
mcandeia, line 46: import { logger } and use it

runtime/caches/redis.ts
```diff
@@ -47,13 +39,33 @@ export const caches: CacheStorage = {
...
     throw new Error("Not Implemented");
   },
-  open: (cacheName: string): Promise<Cache> => {
-    if (!redis) {
-      throw new Error(
-        "Redis coult not be used due to the lack of credentials.",
-      );
-    }
-    const requestURLSHA1 = withCacheNamespace(cacheName);
+  open: async (_: string): Promise<Cache> => {
```
mcandeia, line 42: an important note: 

Open receives `cacheName` as a string, this should be used to "shard" your cache, so ideally you need to use to compose your cache keys to avoid namespace keys clash.

also: open can be called multiple times. a common approach is to create a redis at module level var

let redis: ReturnType<typeof createClient> | null = null;

then here you can just use

redis ??= createClient({ url: REDIS_URL })...

in that way you're going to create a single instance even multiple opens were called
```suggestion
  open: async (_: string): Promise<Cache> => {
```

runtime/caches/redis.ts
```diff
@@ -81,125 +89,78 @@ export const caches: CacheStorage = {
...
-            );
-            return undefined;
-          }
+        ensureConnection();
+
+        return requestURLSHA1(request)
+          .then(redis.get)
+          .then((result: string | null) => {
+            if (!result) {
+              console.debug("[cache]", "miss");
```
mcandeia, line 122: use metrics instead

runtime/caches/redis.ts
```diff
@@ -1,36 +1,28 @@
...
-interface ResponseMetadata {
-  body: string;
-  status: number;
-  headers: [string, string][];
-}
+const REDIS_URL = Deno.env.get("REDIS_URL") || "redis://localhost:6379/0";
 
-function base64encode(str: string): string {
-  return btoa(encodeURIComponent(str));
+function serialize(response: Response): Promise<string> {
```
mcandeia, line 10: an interesting perf test to do is to use a string concatenated with special characters and then split it when read from redis, wondering if JSON operations makes it too slow

runtime/caches/redis.ts
```diff
@@ -81,125 +89,78 @@ export const caches: CacheStorage = {
...
+        return requestURLSHA1(request)
+          .then(redis.get)
+          .then((result: string | null) => {
+            if (!result) {
+              console.debug("[cache]", "miss");
+
+              return undefined;
+            }
+
+            console.debug("[cache]", "hit");
```
mcandeia, line 127: remove

runtime/caches/redis.ts
```diff
@@ -81,125 +89,78 @@ export const caches: CacheStorage = {
...
-        _options?: CacheQueryOptions,
-      ): Promise<ReadonlyArray<Response>> => {
-        throw new Error("Not Implemented");
-      },
-      /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Cache/put) */
       put: async (
         request: RequestInfo | URL,
         response: Response,
       ): Promise<void> => {
+        console.debug("[cache]", "put");
```
mcandeia, line 141: remove

runtime/caches/redis.ts
```diff
@@ -81,125 +89,78 @@ export const caches: CacheStorage = {
...
+          .then(([cacheKey, data]) => {
+            const expirationTimestamp = Date.parse(
+              response.headers.get("expires") ?? "",
+            );
+
+            const ttl = expirationTimestamp - Date.now();
+
+            return redis.set(cacheKey, data, { PX: ttl });
+          })
+          .then(() => console.log("[cache]", "store"))
```
mcandeia, line 162: remove

runtime/caches/redis.ts
```diff
@@ -81,125 +89,78 @@ export const caches: CacheStorage = {
...
+          .catch((err) => console.error("[cache]", err));
```
mcandeia, line 163: use logger

mcandeia (changes requested):

runtime/caches/redis.ts
azisaka: I know the default database will always be 0, but I want to make sure it's explicit for everyone.

runtime/caches/redis.ts
```diff
@@ -1,36 +1,37 @@
...
-  body: string;
-  status: number;
-  headers: [string, string][];
-}
+const REDIS_URL = Deno.env.get("REDIS_URL") || "redis://localhost:6379/0";
+
+const redis = createClient({
+  url: REDIS_URL,
+}).on("error", (err: Error) => {
+  logger.error(err);
```
mcandeia, line 14: if you just output the error using `logger.error` so it becomes difficult to find them, you can use the second parameter which is `Record<string, string>` to pass additional info. you can use 
```json
{
type: "cache_error",
engine: "redis",
operation: "createclient"
}
```

same for others

mcandeia (changes requested):

runtime/caches/redis.ts
```diff
@@ -1,36 +1,37 @@
...
-    enableTelemetry: true,
-  })
-  : null;
-
-interface ResponseMetadata {
-  body: string;
-  status: number;
-  headers: [string, string][];
-}
+const REDIS_URL = Deno.env.get("REDIS_URL") || "redis://localhost:6379/0";
```
mcandeia, line 9:
```suggestion
const REDIS_URL = Deno.env.get("LOADER_CACHE_REDIS_URL") || "redis://localhost:6379/0";
```

mcandeia (changes requested):

runtime/caches/redis.ts
```diff
@@ -47,13 +50,19 @@ export const caches: CacheStorage = {
...
+      url: REDIS_URL,
+    }).on("error", () => {
+      if (client.isOpen) {
+        client.disconnect();
+      }
+
+      setTimeout(async () => redis = await client.connect(), 5000);
+    });
+
+    redis ??= await client.connect();
```
mcandeia, line 64: Question:

Since you're creating a new `client` for every `open` call but assigning `client.connect()` to a single instance, the `on("error")` hook will be called for each client you create? I don't think so, right?

Also:

Since javascript is singlethreaded, the only way to avoid concurrency at this line:

`redis ??= await client.connect();`

is removing the keyword `await`, otherwise the runtime will change the thread context and will have a racing condition

So ideally it would be:


module-level funcs
```ts
let redisPromise:
  | null
  | Promise<RedisClientType<RedisModules, RedisFunctions, RedisScripts>> = null;

const connect = (): Promise<RedisClientType<RedisModules, RedisFunctions, RedisScripts>> => {
const client = createClient({
      url: REDIS_URL,
    }).on("error", async () => {
      if (client.isOpen) {
        client.disconnect();
      }
     await redisPromise; // this is a guard to make sure you're not retrying inflight
      const wait5s = new Promise((resolve) => setTimeout(resolve, 5000)); // this will set a timeout of 5s
      redisPromise = wait5s.then(() => client.connect())
    });
    return client.connect();
}
```
```suggestion
    redisPromise ??= connect(); // this avoid racing conditions
    const redis = await redisPromise;
```

mcandeia (changes requested):

runtime/caches/redis.ts
```diff
@@ -0,0 +1,139 @@
...
+  RedisScripts
+>;
+
+let redis: null | RedisConnection = null;
+
+export const isAvailable = Deno.env.has("LOADER_CACHE_REDIS_URL");
+
+function connect(): boolean | Promise<RedisConnection> {
+  if (!isAvailable) {
+    return false;
```
mcandeia, line 31:
```suggestion
    return;
```

mcandeia (approved):

PR #859, opened 2024-10-09
useScript breaking when called in the browser

saymondamasio:
I checked that `Deno` is not called on the client side.

### DIFF
--- a/hooks/useScript.ts
+++ b/hooks/useScript.ts
@@ -6,10 +6,11 @@
  * Use at your own risk.
  */
 
+import { IS_BROWSER } from "$fresh/runtime.ts";
 import { LRUCache } from "npm:lru-cache@10.2.0";
 import { minify as terserMinify } from "npm:terser@5.34.0";
 
-const verbose = !!Deno.env.get("SCRIPT_MINIFICATION_DEBUG");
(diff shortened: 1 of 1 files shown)


viktormarinho:
Some things to note here:

- $fresh/runtime probably should not be imported in this file, since you can use it independently from the fresh setup
- useScript should probably not be used in the browser at all; it is importing terser

matheusgr:
useScript must fail if run in client side. Its a server side only thing to do.

Maybe you can give a warning and don't run it on client side instead.

hooks/useScript.ts
```diff
@@ -6,10 +6,11 @@
  * Use at your own risk.
  */
 
+import { IS_BROWSER } from "$fresh/runtime.ts";
 import { LRUCache } from "npm:lru-cache@10.2.0";
 import { minify as terserMinify } from "npm:terser@5.34.0";
 
-const verbose = !!Deno.env.get("SCRIPT_MINIFICATION_DEBUG");
+const verbose = !IS_BROWSER && !!Deno.env.get("SCRIPT_MINIFICATION_DEBUG");
```
matheusgr, line 13: I believe that this function should not run client side.

Maybe if it is run at clientSide, throw an exception and stop it execution.

matheusgr (changes requested):

PR #860, opened 2024-10-12
Deno 2 support

mcandeia:


engine/schema/schemeable.bench.ts
```diff
@@ -103,13 +103,14 @@ Deno.bench(
...
+// Deno.bench(
+//   "transform to schemeable using denodoc",
+//   { group: "schema_gen" },
+//   async () => {
+//     localStorage.clear();
+//     await Promise.all(testCases.map(({ path, typeName }) => {
+//       return getSchemeableDenoDoc(path, typeName);
+//     }));
+//   },
+// );
```
hugo-ccabral, line 116: isn't better tojust remove it?

PR #864, opened 2024-10-31
fix: windows path checking

vitoUwu:
This PR fixes some path checking on recursive importing that are related to local apps.

Before this fix, when running on windows a lot of classes were removed because the `resolveRecursively` was skipping the local app path since `path.includes(localAppsFolder)` was comparing different type of separators and the import wasn't occurring

image


mcandeia (approved):
Thanks for you contribution!

PR #867, opened 2024-11-05
Fix upload file

saymondamasio:
It was not possible to upload files, as we were not receiving "content-length" in the request header, so the `getParsingStrategy()` method was not returning the correct strategy.


viktormarinho (approved):
lgtm

PR #872, opened 2024-12-01
Fix - Flag

gabrielMatosBoubee:
image

The variant was return the objext instead the right value. This could break the website, so, I fix to return the right type


mcandeia (approved):

mcandeia:
I took a bit to understand what you're trying to do @gabrielMatosBoubee, for the next time: ask gpt to write the description of the PR that would be better then 🤣

PR #876, opened 2024-12-08
Fix(flags): False positive when variant rule is an async function

vitoUwu:
There are scenarios where we have a multivariant (e.g Pages with variants) and `variant.rule` can return a Promise<boolean> even if the matcher module returns a synchronous function, this occurs when I have a matcher in my project and an (asynchronous) middleware in my app `site.ts `, making `variant.rule` asynchronous as well. It doesn't impact matchers that are from other apps like `website` app that have a lot of matchers, but only the local ones

Illustration of my problem.
```
- Page "Home" 
   |- variants.0.home => Promise<boolean> -> when resolved it returns "false"
   \- variants.1.home => boolean -> always "true"

// looping through the flags.variants
index 0 -> variants.rule(ctx) -> Promise<false> -> true // because Promise is a truly value
early return with wrong variant match
```
The `variant.rule` check is always true since Promise<boolean> is a truly value, even if the final value of the promise is false.

This PR makes it possible for us to detect asynchronous functions in the flags block, where the problem was occurring, and if necessary we make sure the promise is resolved so that we can check if the `variant.rule` returns `true`
```
- Page "Home" 
   |- variants.0.home => Promise<boolean> -> when resolved it returns "false"
   \- variants.1.home => boolean -> always "true"

// looping through the flags.variants
index 0 -> variants.rule(ctx) -> Promise<false> -> await for the result -> don't match
index 1 -> variants.rule(ctx) -> return as it's returning "true"
return with the correct match
```

### DIFF
--- a/blocks/flag.ts
+++ b/blocks/flag.ts
@@ -5,6 +5,7 @@ import JsonViewer from "../components/JsonViewer.tsx";
 import type { Block, BlockModule, InstanceOf } from "../engine/block.ts";
 import { isDeferred } from "../engine/core/resolver.ts";
 import { type Device, deviceOf } from "../utils/userAgent.ts";
+import { isAwaitable } from "../utils/mod.ts";
 export type Flag = InstanceOf<typeof flagBlock, "#/root/flags">;
 
 export interface FlagObj<TVariant = unknown> {
@@ -64,7 +65,7 @@ const flagBlock: Block<BlockModule<FlagFunc>> = {
   >(func: {
     default: FlagFunc<TConfig>;
   }) =>
-  ($live: TConfig, { request }: HttpContext) => {
+  async ($live: TConfig, { request }: HttpContext) => {
     const flag = func.default($live);
     let device: Device | null = null;
     const ctx = {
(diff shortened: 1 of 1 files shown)


blocks/flag.ts
```diff
@@ -64,7 +65,7 @@ const flagBlock: Block<BlockModule<FlagFunc>> = {
   >(func: {
     default: FlagFunc<TConfig>;
   }) =>
-  ($live: TConfig, { request }: HttpContext) => {
+  async ($live: TConfig, { request }: HttpContext) => {
```
mcandeia, line 68: this will impact tracings as only async functions are included on tracing. This shouldn't be though but we should keep an eye on it.

blocks/flag.ts
```diff
@@ -75,12 +76,33 @@ const flagBlock: Block<BlockModule<FlagFunc>> = {
...
+        if (typeof variant?.rule !== "function") {
+          continue;
+        }
+        let ruleResult = variant?.rule(ctx);
+        // the rule can sometimes be a promise and we need to await it to check if it's truthy or not
+        if (isAwaitable(ruleResult)) {
+          try {
+            ruleResult = await ruleResult;
+          } catch (_) {
+            // if the rule throws an error, we don't match this variant
```
mcandeia, line 91: Why are you ignoring the errors? shouldn't errors be thrown? couldn't this silently fail in production causing non-debuggable bugs?

blocks/flag.ts
```diff
@@ -75,12 +76,33 @@ const flagBlock: Block<BlockModule<FlagFunc>> = {
...
+
+      for (const variant of flag.variants || []) {
+        if (typeof variant?.rule !== "function") {
+          continue;
+        }
+        let ruleResult = variant?.rule(ctx);
+        // the rule can sometimes be a promise and we need to await it to check if it's truthy or not
+        if (isAwaitable(ruleResult)) {
+          try {
+            ruleResult = await ruleResult;
```
mcandeia, line 89: in the case you provided (when site.ts has an app middleware) this code runs on serial (one await for each matcher) instead of parallel since all matchers will returns a promise, right? Given that, you should instead, parallelize the rule evaluation.

mcandeia (changes requested):

blocks/flag.ts
vitoUwu: yeah, reading this code now it doesn't make sense at all, now it will throw if that's the case

blocks/flag.ts
vitoUwu: not at all, matchers coming from "website" app continue as `sync` functions, but my local ones, that indirectly depends on my async middleware, turns into `async` function

mcandeia: I mean, anyways introducing a middleware will cause all your matchers to be async, so it worth to make parallel instead, wdyt?

vitoUwu: make sense, your proposal is to always use `Promise.all` on this list, right?

```js
Promise.all(flags.variants.map((variant) => variant.rule(ctx)).find((result) => result)
```

PR #882, opened 2025-01-02
[WIP] Make preview SSR only

matheusgr:
Block previews are susceptible to high memory usage and may generate excessive requests, especially when dealing with non-cacheable js resources (particularly on localhost calls).

To mitigate these issues, we avoid hydration. Consequently, the server responds with a concise HTML snippet, and the client refrains from using JavaScript to render the section. Nevertheless, we still display apps global sections and <Head> elements, ensuring that the theme and any other inserted HTML elements are visible in the preview.

This is a minor step towards resolving the block gallery issue. There are still several aspects that require further attention and improvement.


mcandeia (approved):

matheusgr:
Don't merge yet. The path from page previews / normal navigation on env is the same of the previews. We should only enable SSR on block preview.

PR #884, opened 2025-01-16
[FEAT]: create server only access control for loaders and actions

IncognitaDev:
This PR adds the capability of export a access control const `defaultVisibility` that determines if a loader or a action is server only acessible or public accssible. 

A server only accessible loader/action its only callable by ctx.invoke, this prevent exposing private endpoints on webisite calls on client. 

Exemple:

```ts
export const defaultVisibility = 'private'
```

image

```ts
export const defaultVisibility = 'public'
```
image

### DIFF
--- a/blocks/action.ts
+++ b/blocks/action.ts
@@ -21,8 +21,10 @@ const actionBlock: Block<ActionModule> = {
     TProps = any,
   >(
     mod: ActionModule<TProps>,
+    key: string,
   ) => [
-    applyProps(gateKeeper(mod)),
+    gateKeeper(mod.defaultVisibility, key),
+    applyProps(mod),
   ],
   defaultPreview: (result) => {
     return {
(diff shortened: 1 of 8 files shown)


mcandeia (approved):

IncognitaDev:
I add a `visibilityOverride` to deco plugin that allows overrite the visibility of a block. With this you can explicity turn a default private block to act as public or a public to act as a private one.

usage:

```
import { defineConfig } from "$fresh/server.ts";
import { plugins } from "deco/plugins/deco.ts";
import manifest from "./manifest.gen.ts";

export default defineConfig({
  plugins: plugins<typeof manifest>({
    manifest,
    htmx: true,
    visibilityOverrides:  {
      "site/loaders/minicart.ts": "public",
    },
  }),
});

```

mcandeia (approved):

PR #937, opened 2025-07-04
Use deployment id on render

matheusgr:


runtime/routes/render.tsx
```diff
@@ -77,6 +82,14 @@ export const handler = createHandler(async (
 ) => {
   const { req: { raw: req }, var: state, render } = ctx;
   const opts = fromRequest(req);
+  if (opts.deploymentId && deploymentId) {
+    if (opts.deploymentId !== deploymentId) {
+      throw badRequest({
```
lucis, line 87: Is this necessary? What do you think about logging the error but not failing the request?

lucis (approved):

runtime/routes/render.tsx
```diff
@@ -7,6 +7,8 @@ import { createHandler, DEBUG_QS } from "../middleware.ts";
 import type { PageParams } from "../mod.ts";
 import Render, { type PageData } from "./entrypoint.tsx";
 
+const deploymentId = Deno.env.get("DENO_DEPLOYMENT_ID");
```
hugo-ccabral, line 10: Will it exists in all platforms?

runtime/routes/render.tsx
hugo-ccabral: Perhaps we need to fail the request so a second one can cache the right response. If we dont fail will the "not failure" request be cached?

runtime/routes/render.tsx
igorbrasileiro: This already exists in context, shouldn't import it from context?

hugo-ccabral (approved):

runtime/routes/render.tsx
guitavano: what if we return max-age 0?

runtime/routes/render.tsx
matheusgr: This is not on the fresh context (what is used there).

runtime/routes/render.tsx
matheusgr: We already have this issue with assets. If we request an asset on a older deploy, it will give an 404.

It is better to fail fast and let the CDN cache on a successful response, and try to fix this issue with another approach.

PR #978, opened 2025-10-15
 feat: add configurable CACHE_TTL environment variable

pbpeterson:
## Summary
  - Adds `LRU_CACHE_TTL ` environment variable support to allow manual configuration of cache time-to-live
  - When `LRU_CACHE_TTL ` is set, it takes precedence over the expiration timestamp from response headers
  - Provides more control over cache duration for different deployment scenarios

  ## Changes
  - Added `LRU_CACHE_TTL ` environment variable in `runtime/caches/lrucache.ts:19`
  - Updated cache put logic (lines 76-78) to check for `LRU_CACHE_TTL ` first before falling back to response header expiration
  - Added documentation comment explaining the variable's purpose

  ## Example Scenario

  Consider the following configuration:
  - `LRU_CACHE_TTL ` is set to 3600000 (1 hour)
  - Response from server has `Expires` header indicating 3 minutes from now

  ### Timeline:

  **T=0: Initial Request**
  - Resource is fetched from the server
  - LRU cache stores the entry with TTL of 1 hour (from `LRU_CACHE_TTL `)
  - CacheStorage API stores the response with expiration from headers (3 minutes)

  **T=2 minutes: Second Request (within 3 minutes)**
  - LRU cache check: ✓ Hit (58 minutes remaining)
  - CacheStorage API check: ✓ Fresh (1 minute remaining)
  - Result: Cached response served immediately, no network request

  **T=4 minutes: Third Request (after 3 minutes, within 1 hour)**
  - LRU cache check: ✓ Hit (56 minutes remaining)
  - CacheStorage API check: ✗ Expired (1 minute past expiration)
  - Result: Stale response served immediately to the user
  - Background revalidation: Fresh copy fetched from server and cache updated

  **T=65 minutes: Fourth Request (after 1 hour)**
  - LRU cache check: ✗ Miss (TTL expired)
  - CacheStorage API check: ✗ Miss
  - Result: Fresh fetch from server, cache repopulated

  This dual-layer approach allows you to:
  - Control memory retention independently via `LRU_CACHE_TTL ` (prevents premature eviction from LRU)
  - Respect server-provided expiration for freshness semantics

  This describes how the LRU cache's TTL acts as a retention policy while the CacheStorage API's expiration determines freshness,
   enabling the stale-while-revalidate pattern.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable "stale TTL" increment to optionally extend how long cached items remain valid beyond response expiration times.
  * If set to a valid number, the increment is added to each response's TTL; if unset/invalid, caching follows response expiration as before.
  * Existing expiration-based staleness detection remains unchanged.

### DIFF
--- a/runtime/caches/lrucache.ts
+++ b/runtime/caches/lrucache.ts
@@ -81,9 +81,7 @@ function createLruCacheStorage(cacheStorageInner: CacheStorage): CacheStorage {
           // - If STALE_TTL_PERIOD is configured, add it to the expiration time from the response headers
           //   This allows extending the cache lifetime beyond what the server specifies and serves stale content during this extra time
           // The staleness is detect at the loader level because it checks for the expires header, that remains untouched here, the idea is to serve stale content but with expired header
-          const ttl = !Number.isNaN(STALE_TTL_PERIOD)
-            ? (expirationTimestamp - Date.now()) + STALE_TTL_PERIOD
-            : expirationTimestamp - Date.now();
+          const ttl = (expirationTimestamp - Date.now()) + STALE_TTL_PERIOD;
 
           const cacheKey = await requestURLSHA1(request);
           const length = response.headers.get("Content-Length");


igorbrasileiro:
This PR seems to be related to this https://github.com/deco-cx/deco/pull/977 And the CACHE_TTL can be setup with the CACHE_TTL_RESOLUTION with stale

pbpeterson:
This PR is kinda different.

 This PR (CACHE_TTL):
  - Controls the LRU cache item TTL (how long items stay in the in-memory LRU cache)
  - Does NOT affect the ttlResolution parameter (line 26 in runtime/caches/lrucache.ts)
  - Does NOT change the loader cache duration/behavior
  - Only overrides the TTL for individual cache entries when storing them in the LRU cache (lines 79-81)

  Why we keep CACHE_TTL_RESOLUTION unchanged:
  We need the precision of CACHE_TTL_RESOLUTION (default 30 seconds) for the background task that checks and purges expired items
   from the LRU cache. This is independent of how long individual items should be cached.

  The separation allows:
  - CACHE_TTL: Set how long cache entries are valid (e.g., 5 minutes = 300000ms)
  - CACHE_TTL_RESOLUTION: Keep precise control over how frequently the cache checks for expired items (e.g., every 30 seconds)

runtime/caches/lrucache.ts
```diff
@@ -16,6 +16,8 @@ const CACHE_ALLOW_STALE = Deno.env.get("CACHE_ALLOW_STALE") !== "false"; // auto
 const CACHE_TTL_RESOLUTION = parseInt(
   Deno.env.get("CACHE_TTL_RESOLUTION") ?? "30000",
 ); // check for expired items every 30 seconds
+// Time-to-live in milliseconds for cached items. If not set, uses the expiration timestamp from response headers
+const CACHE_TTL = Deno.env.get("CACHE_TTL");
```
hugo-ccabral: Please use parseInt here so not do it everytime you call a cache operation

runtime/caches/lrucache.ts
```diff
@@ -72,7 +74,11 @@ function createLruCacheStorage(cacheStorageInner: CacheStorage): CacheStorage {
             response.headers.get("expires") ?? "",
           );
 
-          const ttl = expirationTimestamp - Date.now();
+          const cacheTtlParsed = parseInt(CACHE_TTL ?? "");
```
hugo-ccabral: dont do the parseInt everytime

runtime/caches/lrucache.ts
```diff
@@ -72,7 +74,11 @@ function createLruCacheStorage(cacheStorageInner: CacheStorage): CacheStorage {
...
+
+          const ttl = !Number.isNaN(cacheTtlParsed)
+            ? cacheTtlParsed
```
hugo-ccabral: perhaps it is better to do this at the high level cache engine. 
your proposal is acceptable but only apply to this cache behavior and we use tiered cache with support to multiple engines. 
Currently we only have this one but if in the future we decide to have another one your ttl would only be applied here

hugo-ccabral (changes requested):

runtime/caches/lrucache.ts
pbpeterson: This works a bit differently — I'm modifying the TTL of entries in the LRU Cache that map to keys in the Cache API. It effectively enables a stale-while-revalidate mechanism with a configurable time window.

pbpeterson: Would it be better to change the env name to `LRU_CACHE_TTL`?

runtime/caches/lrucache.ts
```diff
@@ -16,6 +16,11 @@ const CACHE_ALLOW_STALE = Deno.env.get("CACHE_ALLOW_STALE") !== "false"; // auto
 const CACHE_TTL_RESOLUTION = parseInt(
   Deno.env.get("CACHE_TTL_RESOLUTION") ?? "30000",
 ); // check for expired items every 30 seconds
+// Additional time-to-live increment in milliseconds to extend the cache expiration beyond the response's Expires header.
+// If not set, the cache will use only the expiration timestamp from response headers
+const CACHE_TTL_INCREMENT = parseInt(
```
hugo-ccabral: what about STALE_TTL_PERIOD

runtime/caches/lrucache.ts
```diff
@@ -72,7 +77,14 @@ function createLruCacheStorage(cacheStorageInner: CacheStorage): CacheStorage {
             response.headers.get("expires") ?? "",
           );
 
-          const ttl = expirationTimestamp - Date.now();
+          // Calculate the time-to-live (TTL) for the cached item:
+          // - If CACHE_TTL_INCREMENT is configured, add it to the expiration time from the response headers
+          //   This allows extending the cache lifetime beyond what the server specifies
```
hugo-ccabral:
```suggestion
          // - If STALE_TTL_PERIOD is configured, add it to the expiration time from the response headers
          //   This allows extending the cache lifetime beyond what the server specifies and serves stale content during this extra time
          // The staleness is detect at the loader level because it checks for the expires header, that remains untouched here, the idea is to serve stale content but with expired header
```

hugo-ccabral (changes requested):

PR #1004, opened 2025-11-17
remove revision ID from loaders cache key

guitavano:
Created to reuse cache across different revisions.

## Summary by CodeRabbit

* **Refactor**
  * Simplified loader flow and streamlined cache-key construction to prefer an environment-provided revision value with a safe fallback.

* **Bug Fixes**
  * If resolution chain or revision value is missing, caching is skipped, a warning is logged, timing is preserved, and requests are forwarded to the underlying handler.

### DIFF
--- a/blocks/loader.ts
+++ b/blocks/loader.ts
@@ -5,6 +5,7 @@ import { ValueType } from "../deps.ts";
 import type { Block, BlockModule, InstanceOf } from "../engine/block.ts";
 import { FieldResolver } from "../engine/core/resolver.ts";
 import { singleFlight } from "../engine/core/utils.ts";
+import type { DecofileProvider } from "../engine/decofile/provider.ts";
 import { HttpError } from "../engine/errors.ts";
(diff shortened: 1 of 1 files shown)


blocks/loader.ts
```diff
@@ -262,34 +260,28 @@ const wrapLoader = (
...
-            );
-          }
+        if (!resolveChainString) {
+          logger.warn(`Could not get resolveChain`);
+          timing?.end();
+          return await handler(props, req, ctx);
+        }
 
+        // Optimize cache key generation using simple string concatenation
+        const kRevision = Deno.env.get("K_REVISION") ?? "";
```
hugo-ccabral: Why don't you use the previous revision definition as fallback?
```suggestion
        const kRevision = Deno.env.get("K_REVISION") ?? (await release?.revision() ?? undefined);
```

hugo-ccabral: otherwise it wont cache at all if k_revision is missing, like locally

PR #1185, opened 2026-05-11
feat(daemon): persist env workspace to deco/env/<name> branch

hugo-ccabral:
## Summary

Adds a git-backed persistence path for env pods so admin can drop the EBS PVC that today stores the cloned repo + WIP across restarts. Paired with [deco-sites/admin#3160](https://github.com/deco-sites/admin/pull/3160) which removes the PVC from the StatefulSet manifest.

When `DECO_ENV_NAME` is set:
- **`ensureGit`** targets ``deco/env/<DECO_ENV_NAME>``. It probes with `ls-remote` first; if the branch exists, clones it directly; otherwise clones the upstream branch and `checkout -b`'s the env branch locally. Upstream is wired to `origin/<upstream>` so `assertRebased` rebases env state onto main on startup.
- **`flushEnvBranch()`** commits + pushes the working tree to the env branch. Write-locked against publish/rebase/file-watch, skips when the tree is clean, refreshes the GitHub App token, never throws.
- **`daemon/main.ts`** wires `flushEnvBranch` on a 2-min throttle in the file watcher, and into a `SIGTERM`/`SIGINT` handler capped at 60s so it cannot run past the pod's `terminationGracePeriodSeconds`.

Production pods (no `DECO_ENV_NAME`) are unaffected — every new code path is gated on `ENV_BRANCH` being set, and the existing `SOURCE_ASSET_PATH` tarball persistence continues to operate for post-publish snapshots.

## Why

Env pods today rely on an EBS PVC to keep `.git/` + the user's uncommitted edits across scale-to-zero cycles. EBS attach dominates cold-start (~30–60s), pins the pod to a specific AZ, and complicates pod scheduling. Pushing WIP to a side branch on the user's own repo lets us throw the disk away on shutdown and clone fresh on startup — same persistence outcome, no attach delay, no AZ pinning.

## Test plan

- [ ] Run a local env pod against a throwaway fork with `DECO_ENV_NAME=test-1`; verify the branch `deco/env/test-1` appears on the fork after edits + restart.
- [ ] Restart cycle: verify the env branch is rehydrated on cold start (clone visible in `[step 1/4]` log).
- [ ] Push to upstream `main` while the env is running, scale env down + up, confirm env branch came back rebased onto the new upstream.
- [ ] Force-kill the pod (`kill -9`) and verify the next cold start picks up the last successful periodic flush (loss bounded to ~2 min of edits).
- [ ] Sandbox mode (no `DECO_ENV_NAME`): confirm unchanged behavior — no env branch, no flush, no SIGTERM handler installed.

### DIFF
--- a/daemon/git.ts
+++ b/daemon/git.ts
@@ -21,6 +21,13 @@ const GITHUB_APP_KEY = Deno.env.get("GITHUB_APP_KEY");
 const BUILD_FILES_DIR = Deno.env.get("BUILD_FILES_DIR");
 const ADMIN_DOMAIN = "https://admin.deco.cx";
 
+const envBranchName = (): string | undefined => {
+  const envName = Deno.env.get("DECO_ENV_NAME");
+  return envName ? `deco/env/${envName}` : undefined;
+};
+
+export const ENV_BRANCH = envBranchName();
+
 /**
  * Parse owner and repo name from a GitHub URL.
  * Handles HTTPS (https://github.com/owner/repo.git) and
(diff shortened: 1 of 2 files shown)


daemon/git.ts
```diff
@@ -794,17 +803,72 @@ export const ensureGit = async ({
         ? `https://github.com/deco-sites/${site}.git`
         : `git@github.com:deco-sites/${site}.git`);
```
o-pauloroberto, line 804: Colocar a ORG do github como um parâmetro para não ficar hardcoded

PR #1206, opened 2026-06-01
fix(daemon): robust sandbox cold-start (eager warm + fast-503 + optional warmup)

nicacioliveira:
## Problem

Sandbox/preview environments (`eks-envs`) intermittently return **504** to the user even though the pod is healthy. Concrete case: `envs-farmrio--s3a7o0.decocdn.com` — slow to open, then 504 before finishing; the nginx metric shows **499** (client closed the connection).

### Diagnosis

- The `eks-envs` nginx has `proxy-read-timeout: 3600` (intentional, for SSE/HMR and the daemon `/watch` endpoints), so **nginx is not the one timing out**. The cut happens at **Cloudflare** (`decocdn.com`) on the origin timeout (~100s) → **504** to the browser, **499** at the ingress.
- The env is served in **dev mode**; on a cold/idle pod the cold-start work is paid **inline on the first request**, and the daemon **holds the connection open** for the whole boot (`worker()` → `waitUntilReady`, default 60s), so a slow boot/render runs straight past the CDN timeout.
- For a plain env deploy, `currentSite.ensureStarted()` (git clone, build-cache download, manifest + blocks generation — and, via `watchMeta`, the dev-server boot) was only triggered when the deploy carried an AI task. Otherwise it ran lazily on the first request.

## Change — three coordinated pieces (all `SANDBOX_MODE`-scoped)

1. **Eager warm on deploy** — hoist `currentSite.ensureStarted()` to run unconditionally right after the site is assigned. This moves git clone / cache / manifest **and the dev-server boot** (`watchMeta` calls `worker()`) off the first request's critical path. Idempotent; the AI-task path is unchanged.

2. **Fast-503 while the worker boots** (`worker.ts`) — instead of holding the request open for the entire cold boot (which lets the CDN time out as a 504), reply **503** after 5s. The env ingress (`error_page 502 503 -> @admin`) then bounces the client to the activator and retries; the boot keeps running in the background, so the retry lands on a warm worker. **Normal (non-sandbox) mode keeps awaiting `worker()` as before** — no behavior change outside sandboxes.

3. **Optional deep warmup** (`DECO_SANDBOX_WARMUP=true`, **off by default**) — after git clone, fire one internal `GET /` so the dev server JIT-compiles and renders the entry route before the first real request. Carries an `x-deco-warmup` header so the site can skip side effects. Off by default because it triggers one synthetic homepage render (and any server-side analytics it fires).

### How they combine

```
deploy/claim ─▶ ensureStarted()  → deps + dev-server boot start eagerly
              ─▶ (optional) warmup GET / → entry route compiled before user
first request ─▶ if worker not ready in 5s → 503 → activator bounce → retry warm
```

(1)+(2) make the common case robust without holding connections open; (3) closes the remaining first-render gap when enabled.

## Caveats / follow-ups

- Even with all three, a deep uncached route whose legitimate render exceeds ~100s (e.g. slow VTEX upstream) can still time out — that's an app/upstream concern (cache/SWR), not infra cold-start.
- Requires an image release of `ghcr.io/deco-cx/deco/ai` and a tag bump in `infra_applications` (the `eks-envs` SandboxTemplate) to take effect in production. `DECO_SANDBOX_WARMUP` would be set there to enable piece 3.

### Ruled out during diagnosis
- A readinessProbe on the SandboxTemplate would break the warmpool — free (unclaimed) pods serve a legitimate 503 until claimed, so the pool would never reach `readyReplicas`.
- Lowering `proxy-read-timeout` would break SSE/HMR (it is 3600 on purpose).

## Summary by CodeRabbit

* **New Features**
  * Optional sandbox warmup that issues an internal request at startup to prepare JIT/first-render and logs success/failure.

* **Improvements**
  * Eager startup initialization to run git/setup, manifest and metadata generation, and start file watching sooner.
  * Sandbox readiness gate that returns HTTP 503 with retry-after when startup is delayed; warmup requests bypass the gate and continue to wait.

* **Chores**
  * Benchmark import adjusted to avoid type-check issues on fresh fetch.

### DIFF
--- a/blocks/loader.bench.ts
+++ b/blocks/loader.bench.ts
@@ -1,4 +1,8 @@
-import hash from "https://esm.sh/v135/object-hash@3.0.0";
+// `?no-dts` skips esm.sh's bundled type definitions. Upstream `@types/object-hash`
+// uses a CommonJS `export =` with no synthetic default, which makes Deno's
+// type-check fail with TS1192 ("no default export") on a fresh fetch. The types
+// aren't needed in this benchmark, so dropping them keeps CI deterministic.
+import hash from "https://esm.sh/v135/object-hash@3.0.0?no-dts";
 
 const props = {
   randomObject: {
--- a/daemon/main.ts
+++ b/daemon/main.ts
@@ -38,6 +38,8 @@ import { register, type TunnelConnection } from "./tunnel.ts";
 import {
   createWorker,
   resetWorkerState,
+  WARMUP_HEADER,
+  WARMUP_TOKEN,
   worker,
   type WorkerOptions,
 } from "./worker.ts";
@@ -619,15 +621,17 @@ if (SANDBOX_MODE) {
       // entry route before the first real user request arrives — otherwise that
       // compile + render is paid inline on the user's first hit. Off by default
       // because it triggers one synthetic homepage render (and any server-side
-      // analytics it fires); the x-deco-warmup header lets the site opt out of
-      // side effects. Enable with DECO_SANDBOX_WARMUP=true.
+      // analytics it fires); the x-deco-warmup header (carrying the per-process
+      // token) marks it as our internal warmup so it bypasses the fast-503 gate
+      // and the site can opt out of side effects. Enable with
+      // DECO_SANDBOX_WARMUP=true.
       if (Deno.env.get("DECO_SANDBOX_WARMUP") === "true") {
         const { app: siteApp, gitReady } = currentSite;
         gitReady
           .then(() =>
             siteApp.fetch(
               new Request("http://localhost/", {
-                headers: { "x-deco-warmup": "1" },
+                headers: { [WARMUP_HEADER]: WARMUP_TOKEN },
               }),
             )
           )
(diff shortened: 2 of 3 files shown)


guitavano (approved):

nicacioliveira:
Addressed the bot review findings in 6885b712:

- **cubic (worker.ts:112) — forgeable `x-deco-warmup` header:** the warmup bypass now requires a per-process random token (`WARMUP_TOKEN`) carried only by the daemon's own internal warmup request. A forged header from public traffic won't match and is treated as a normal request.
- **cubic (worker.ts:130) — `isWorkerDisabled()` only covers one failure class:** restructured the readiness race so that **any** `worker()` rejection (missing `dev.ts`, crash, etc.) → **424** (non-retryable), and only a still-pending boot at the gate → **503** (retryable, bounces to the activator). The rejection is mapped inline so the race loser can't surface as an unhandled rejection after we reply.
- **CodeRabbit / cubic (worker.ts:124) — permanent failures returning transient 503:** covered by the same restructure (rejection → 424).

`deno fmt`, `lint`, and `check` pass on both files.

PR #1234, opened 2026-09-16
fix(engine): dispose superseded ReleaseResolver's release subscription

nicacioliveira:
## Problem

Every `new ReleaseResolver` subscribes to its release in the constructor, and the returned `Disposable` is dropped:

```ts
this.release.onChange(() => { /* clears hints, caches */ });   // Disposable discarded
```

`installApps` rebuilds the resolver on every decofile change:

```ts
currentResolver = currentResolver.with({ resolvers, resolvables });
```

`.with()` constructs a new `ReleaseResolver`, which subscribes to the **same** provider. The superseded resolver is still referenced by the provider's listener list, so it can never be collected — along with its entire `resolvables` / `resolvers` / `resolveHints` graph. One leaked resolver per rebuild, forever.

Since `provider.onChange(...)` triggers `installApps()`, this scales with **publishes**, not with traffic — which is why it shows up as memory that grows with wall-clock time on busy sites and is invisible under load testing.

## Measurement

Reproduced on a production replica — same runner image, Deno 1.44.4, real decofile from a live site — with `heapUsed` read after two forced `HeapProfiler.collectGarbage` calls:

| reload | heapUsed |
|---|---|
| 0 (baseline) | 120,0 MB |
| 2 | 142,8 MB |
| 4 | 153,3 MB |
| 6 | 166,2 MB |
| 8 | 183,5 MB |
| 10 | 204,8 MB |

**~8,5 MB retained per publish**, linear, not reclaimable by GC. `external` stayed flat throughout, so it is heap, not buffers. Repeating the series with concurrent request load gave 8,8 MB/reload — load does not amplify it.

On the site that surfaced this (24–36 publishes/day), a 44h-old pod held ~900 MB of heap after a forced GC, against ~103 MB on a freshly started one.

## Fix

Keep the `Disposable`, expose an idempotent `dispose()` (also wired to `Symbol.dispose`), and drop the superseded resolver's subscription in `installApps`.

Disposing a superseded resolver is safe: in-flight requests holding a reference keep working, they just stop being notified about release changes they no longer serve.

## Test

`dispose() unsubscribes the resolver from its release` pins the subscription count across 25 rebuilds. It is written to stay meaningful **without** the fix — `dispose` is accessed through a cast, so on `main` it compiles, the subscriptions pile up and the assertion reports the actual leak instead of failing to type-check:

```
without fix: AssertionError: Values are not equal: each publish must leave
             exactly one live subscription, not accumulate one per rebuild
with fix:    ok | 4 passed (13 steps) | 0 failed
```

`deno lint` and `deno fmt --check` clean on the touched files.

## Not covered

`runtime/mod.ts` and `runtime/routes/blockPreview.tsx` also call `.with()` per request, but both pass a fresh `release` (the draft provider), so the subscription lands on a short-lived object and goes away with it. Left untouched.

---
## Summary by cubic
Fixes a memory leak in `ReleaseResolver` where every decofile change left the superseded resolver subscribed to its release provider, retaining its whole resolver graph (~8.5 MB per publish). Disposal now works end-to-end: the resolver keeps its subscription disposable, `fromEndpoint` forwards the inner subscription, and `fulfillContext` disposes the outgoing runtime resolver once its replacement is live.

**Bug Fixes**
- `dispose()` is idempotent and wired to `Symbol.dispose`; all four `onChange` providers guard double disposal so it can't drop a live callback.
- In-flight requests on a superseded resolver keep working; they just stop being notified of release changes.
- `fromEndpoint` proxies the inner subscription, so disposal actually unsubscribes in production instead of returning a no-op.
- `fulfillContext` disposes the installed resolver after its replacement is live; the zero-apps path disposes the resolver it discarded instead of the live one.
- The test rebuilds a resolver 25 times and asserts one live subscription, covering fake and endpoint-backed providers.
- `runtime/mod.ts` and `runtime/routes/blockPreview.tsx` call `.with()` with a fresh release, so they are unaffected.

<sup>Written for commit 7f09331592093aeed0eeb9816b38ac1313a8ad1d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/deco-cx/deco/pull/1234?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed repeated application updates leaving behind stale resolver subscriptions.
  * Improved cleanup when replacing resolvers, preventing obsolete resolver state from remaining active.
  * Made resolver and change-listener disposal safe to call multiple times.
  * Ensured cleanup works correctly even when providers have not finished loading.
  * Prevented duplicate or unintended removal of live subscriptions during cleanup.
  * Added automatic cleanup when resolver instances are disposed.


hugo-ccabral (approved):

engine/core/mod.ts
```diff
@@ -101,14 +108,37 @@ export class ReleaseResolver<TContext extends BaseContext = BaseContext> {
     this.danglingRecover = config.danglingRecover;
     this.resolveHints = hints ?? {};
     this.runOncePerRelease = oncePerRelease ?? {};
-    this.release.onChange(() => {
+    this.releaseSubscription = this.release.onChange(() => {
```
hugo-ccabral, line 111: **CRITICAL — the fix is a no-op for essentially every production `DecofileProvider`.**

`this.releaseSubscription` is only real if `release.onChange(cb)` returns a `Disposable` whose `[Symbol.dispose]` actually unsubscribes. It does not, for the provider type used by every real deployment:

```ts
// engine/decofile/fetcher.ts:238-244 (fromEndpoint, unchanged by this PR)
onChange: (cb) => {
  decofileProviderPromise.then((r) => r.onChange(cb));
  return {
    [Symbol.dispose]: () => {},   // discards the real inner Disposable, no-ops
  };
},
```

`getProvider()` (`engine/decofile/provider.ts`) wraps every `folder://`, `file:///app/decofile/*`, `deconfig://` and `http(s)://` release through `fromEndpoint`, and for the common single-provider case `compose(...providers)`'s `.reduce()` (no initial value, 1-element array) returns that object unchanged — so `ReleaseResolver.dispose()` calls straight into this no-op. The subscription this PR is trying to release stays registered on the real inner provider (`engine/decofile/fs.ts` / `fsFolder.ts`) forever, regardless of this fix.

The PR's own production-replica measurement almost certainly went through `fromEndpoint`, so this fix likely does not move the needle on the reported heap growth at all. `fromEndpoint`'s `onChange` needs to await `decofileProviderPromise` and return the *real* inner `Disposable`, e.g.:

```ts
onChange: (cb) => {
  let disposed = false;
  let inner: Disposable | undefined;
  decofileProviderPromise.then((r) => {
    if (disposed) return;
    inner = r.onChange(cb);
  });
  return { [Symbol.dispose]: () => { disposed = true; inner?.[Symbol.dispose](); } };
},
```

engine/manifest/manifest.ts
```diff
@@ -194,7 +194,12 @@ const installAppsForResolver = async (
       ...importMap,
       imports: { ...appImportMap?.imports, ...importMap?.imports },
     };
+    const supersededResolver = currentResolver;
     currentResolver = currentResolver.with({ resolvers, resolvables });
+    // Drop the superseded resolver's release subscription; otherwise every
+    // publish leaves one more resolver graph permanently reachable from the
+    // provider's listener list.
+    supersededResolver.dispose?.();
```
hugo-ccabral: **CRITICAL — this disposes resolvers superseded *within* one `installApps()` run, but not the resolver superseded *across* publishes, which is the PR's actual stated scenario.**

`fulfillContext` (this same file, ~line 297-373) creates one stable `resolver` once, then re-runs `installApps()` on every `provider.onChange` — i.e. every site publish:

```ts
const resolver = new ReleaseResolver<RouteContext>({ resolvers: defaultResolvers, release: provider }); // line 297-300, created once
const installApps = async () => {
  ...
  const { resolver: currentResolver, ... } = await installAppsForResolver(
    resolver.with({ resolvers, danglingRecover }),   // line 321: ALWAYS rebuilt from the same stable base
    newManifest, currentImportMap,
  );
  ...
  ctx.runtime = Promise.resolve(runtime);   // line 369: plain overwrite, no dispose of the outgoing resolver
};
provider.onChange(() => { ...; installApps()...; });   // line 373: fires once per publish
```

`installAppsForResolver`'s do-while loop (this hunk) only disposes resolvers it creates and replaces *inside a single call*. Its **final** `currentResolver` — the one that becomes `ctx.runtime.resolver` and actually serves traffic — is returned undisposed by design (it needs to stay live). But on the *next* publish, `installApps()` builds an entirely new chain from the same base `resolver` (not from the previous `ctx.runtime.resolver`), and line 369 just overwrites `ctx.runtime` — nothing calls `.dispose()` on the resolver that was just superseded. Same result when `apps.length === 0` (line ~331-339): that cycle's resolver is returned via `runtimePromise.resolve(...)` with no reference retained and no dispose either.

Net effect: this PR reduces the leak from *N resolvers per publish* (every intermediate `.with()` in the do-while loop) down to *1 resolver per publish* (the final one, handed off between cycles) — but doesn't eliminate it. That 1-per-publish leak is exactly "since `provider.onChange(...)` triggers `installApps()`, this scales with publishes" from the PR description, still present after this fix. It also keeps the orphaned resolver's own `onChange` callback (`dispatchEvent(new Event("deco:hmr"))`, clearing its own dead caches) firing on every future publish — unbounded CPU on top of the retained memory.

Suggest tracking the previous cycle's resolver (e.g. a `let previousResolver` in `fulfillContext`) and disposing it right after the new one is installed into `ctx.runtime`/`runtimePromise`, on both the "apps installed" and "zero apps" exit paths.

(Minor aside: `supersededResolver.dispose?.()` here uses optional chaining, but `dispose` is a required public field on `ReleaseResolver<RouteContext>` per its type — `?.` is dead defensiveness, `.dispose()` would do.)

engine/core/mod.test.ts
```diff
@@ -50,6 +53,51 @@ Deno.test(".with({ release }) does not inherit stale resolve hints", async () =>
...
+Deno.test("dispose() unsubscribes the resolver from its release", () => {
+  // `installApps` rebuilds the resolver on every decofile change:
+  //   currentResolver = currentResolver.with({ resolvers, resolvables })
+  // Each `new ReleaseResolver` subscribes to the SAME provider. Without
+  // disposing the superseded resolver, its subscription keeps it — and its
+  // whole resolvables/resolvers/resolveHints graph — reachable from the
+  // provider's listener list forever, so every publish permanently retains one
+  // more resolver. This test pins the subscription count instead of the bytes.
+  const listeners = new Set<() => unknown>();
+  const release = {
```
hugo-ccabral, line 65: **Test coverage gap — this only proves the mechanism works against a hand-rolled fake, not against any real `DecofileProvider`.**

The fake `release` here implements `onChange` correctly and synchronously:

```ts
onChange: (cb) => { listeners.add(cb); return { [Symbol.dispose]: () => listeners.delete(cb) }; },
```

but the real production provider (`fromEndpoint` in `engine/decofile/fetcher.ts:238-244`) does not — it discards the inner `Disposable` and returns a no-op `[Symbol.dispose]`, so `dispose()` never actually unsubscribes for a real site (see the comment on `engine/core/mod.ts:111`). This test is green either way, so CI cannot catch that gap. Consider adding a variant of this test that runs `dispose()` through `fromEndpoint(...)` (or at least `fromJSON(...)` composed the way `getProvider()` composes it) instead of only the fake, so the assertion is meaningful against the actual provider implementations this fix is meant to work with.

engine/core/mod.ts
```diff
@@ -101,14 +108,37 @@ export class ReleaseResolver<TContext extends BaseContext = BaseContext> {
...
+   * dropping the previous subscription the superseded resolver is still
+   * referenced by the provider's listener list and can never be collected, so
+   * every publish permanently retains one full resolver graph.
+   *
+   * Safe to call on a superseded resolver: in-flight requests holding a
+   * reference keep working, they simply stop being notified of release changes
+   * they no longer serve. Idempotent.
+   */
+  public dispose = (): void => {
+    this.releaseSubscription?.[Symbol.dispose]?.();
```
hugo-ccabral: **Latent double-dispose footgun in the real `onChange` implementations this now actually calls.**

`this.releaseSubscription?.[Symbol.dispose]?.()` is the first real caller (in production code) of the `Disposable`s returned by `engine/decofile/fs.ts:157`, `fsFolder.ts:220`, `fetcher.ts`'s `fromJSON:153`, and `realtime.ts:160`, all of which implement disposal as:

```ts
[Symbol.dispose]: () => { onChangeCbs.splice(onChangeCbs.indexOf(cb), 1); },
```

If this ever runs twice for the same subscription (`indexOf` returns `-1` on the second call), `splice(-1, 1)` silently deletes the *last* element of the array — an unrelated, still-active listener — instead of being a no-op. `ReleaseResolver.dispose()` itself guards against calling it twice on the same instance (it nulls `releaseSubscription` after use), so this isn't reachable through today's call sites, but it's a sharp edge now that disposal is actually exercised: worth a guard (`if (idx !== -1) onChangeCbs.splice(idx, 1);`) in all four implementations rather than relying on every caller to stay perfectly idempotent forever.

Also: the second `?.` here is dead — `Disposable`'s `[Symbol.dispose]` is a required member of the type, so once `this.releaseSubscription` is confirmed non-undefined, `[Symbol.dispose]` is guaranteed to exist. `this.releaseSubscription?.[Symbol.dispose]()` says the same thing with one less optional-chain.

engine/core/mod.ts
```diff
@@ -101,14 +108,37 @@ export class ReleaseResolver<TContext extends BaseContext = BaseContext> {
...
+    this.releaseSubscription = undefined;
+  };
+
+  [Symbol.dispose](): void {
```
hugo-ccabral: **Minor: `[Symbol.dispose]` is currently dead API surface with a `this`-binding trap, and duplicates `dispose` above it.**

`dispose` is a bound arrow class field; `[Symbol.dispose]` is a plain prototype method that forwards to it. Every call site today (`x.dispose?.()`, or an implicit `using x = ...` invoking it as `x[Symbol.dispose]()`) keeps `this` bound correctly, so nothing breaks *today*. But there's no `using` declaration anywhere in the codebase for `ReleaseResolver` — grepping the worktree turns up none — so this method exists purely for a consumption pattern nothing uses yet, and if it's ever destructured or passed as a bare reference (`const d = resolver[Symbol.dispose]; d()`), `this` is `undefined` inside `dispose()` in strict mode and it throws.

Could be a one-line field alias instead of a separate method + indirection:

```ts
readonly [Symbol.dispose] = this.dispose;
```

